PR #17, opened 2019-09-27
Replace nixInstallerSha256 with fixed sha256's

Gerschtli:


t184256:
Builds as expected, thanks!

PR #18, opened 2019-10-04
[WIP] Improve setup

Gerschtli:
Hey,

this is a WIP pull request to gather some initial feedback.

The idea was to install all basic packages on boot via a custom nix-channel.

Steps to complete this PR from my perspective:
- [ ] All files in `/bin` should be linked to `~/.nix-profile`
    - [x] `login`
    - [x] `login-inner`
    - [ ] `proot`
    - [x] `sh`
- [ ] All files in `/etc` should be linked to `~/.nix-profile`
    - [x] `home.nix.default`
    - [x] `nix/nix.conf`
    - [x] `resolv.conf`
    - [x] `group` and `passwd` (optional, as I don't know how to get the current uid via nix)
- [x] All files in `/usr/bin` should be linked to `~/.nix-profile`
    - [x] `env`

Furthermore, I need to refactor the `default.nix` a bit, it's just a plain set of derivations..

Maybe we should improve the log messages with colors or something else, but this wasn't my primary concern.

What do you say to the way I implemented this? Feedback welcome :)


t184256:
I read 3, you push 8 more =)

src/pkgs/files/login-inner.nix
```diff
@@ -16,12 +16,32 @@ writeScript "login-inner" ''
...
-  if [ ! -e $HOME/.nix-profile/etc/ssl/certs/ca-bundle.crt ]; then
-    if [ -e ${packageInfo.cacert} ]; then
-      export NIX_SSL_CERT_FILE=${packageInfo.cacert}
-    fi
+  if [ -e ${instDir}/etc/UNINTIALISED ]; then
+    export NIX_SSL_CERT_FILE=${packageInfo.cacert}
+
+    echo "Installing and updating nix-channels..."
+    ${packageInfo.nix}/bin/nix-channel --add https://nixos.org/channels/nixos-19.03 nixpkgs
+    ${packageInfo.nix}/bin/nix-channel --add https://github.com/t184256/nix-on-droid-bootstrap/archive/master.tar.gz nix-on-droid
```
t184256: I envisioned it as a separate repository for some reason. But it can always be extracted later.

src/pkgs/files/login-inner.nix
```diff
@@ -16,12 +16,32 @@ writeScript "login-inner" ''
...
+    ${packageInfo.nix}/bin/nix-channel --update
+
+    echo "Installing nix-on-droid.basic-environment..."
+    ${packageInfo.nix}/bin/nix-env -iA nix-on-droid.basic-environment
+
+    echo "Setting up static symlinks via nix-on-droid-linker"
```
t184256: I find them more dynamic, actually =)

src/pkgs/files/login-inner.nix
```diff
@@ -16,12 +16,32 @@ writeScript "login-inner" ''
...
+  if [ -e ${instDir}/etc/UNINTIALISED ]; then
+    export NIX_SSL_CERT_FILE=${packageInfo.cacert}
+
+    echo "Installing and updating nix-channels..."
+    ${packageInfo.nix}/bin/nix-channel --add https://nixos.org/channels/nixos-19.03 nixpkgs
+    ${packageInfo.nix}/bin/nix-channel --add https://github.com/t184256/nix-on-droid-bootstrap/archive/master.tar.gz nix-on-droid
+    ${packageInfo.nix}/bin/nix-channel --update
+
+    echo "Installing nix-on-droid.basic-environment..."
+    ${packageInfo.nix}/bin/nix-env -iA nix-on-droid.basic-environment
```
t184256: I'm against imperative PM using nix-env. Please accomplish the same in reverse: drop in the default config here and build the first home-manager generation, which, in turn, would install basic-environment.

t184256:
Wait a minute, I feel like I'm missing something.

You want to have an updateable basic-environment even in scenarios when the user doesn't opt in into h-m?

I feel like it brings more complexity and, at the same time, I don't see many use cases for that. Could you elaborate why you think this is necessary?

What I proposed was simpler. Provide initial links in the zipball, pull in a derivation that would replace them from h-m config. No intermediate stage with an opaque uncontrollable magic thingie stealthily added through nix-env, no worries about discoverability, conflicting with the user wishes, updates...

src/pkgs/files/login-inner.nix
Gerschtli: I think it should be in the same repo as the other config files, because otherwise you would need to duplicate scripts like `bin/login` in to separate repositories.

src/pkgs/files/login-inner.nix
Gerschtli: You are completely right :D

src/pkgs/files/login-inner.nix
Gerschtli: But in that way we would force the installation of home-manager what we not intended to do.

src/pkgs/files/login.nix
```diff
@@ -7,14 +7,20 @@ writeScript "login" ''
   #!/system/bin/sh
   set -e
 
-  export USER=$(/system/bin/whoami)
+  export USER=nix-on-droid
```
t184256: I'm afraid of a scenario where the android user*name* somehow leaks into the nix-on-droid environment and doesn't get mapped to anything meaningful. There was a reason why I did bother with all that stuff, but I didn't track history in git back then, so I cannot exactly recall my reasoning behind that.

src/pkgs/files/login.nix
```diff
@@ -7,14 +7,20 @@ writeScript "login" ''
...
   export PROOT_L2S_DIR=${instDir}/.l2s
 
   if [ ! -e ${instDir}/etc/passwd ]; then
     [ -n "$@" ] || echo "Creating /etc/passwd..."
     echo "root:x:0:0:System administrator:${instDir}/root:/bin/sh" > ${instDir}/etc/passwd
-    echo "nix-on-droid:x:$(/system/bin/stat -c '%u:%g' ${instDir}):nix-on-droid:/data/data/com.termux.nix/files/home:/bin/sh" >> ${instDir}/etc/passwd
+    echo "$USER:x:$(/system/bin/stat -c '%u:%g' ${instDir}):$USER:/data/data/com.termux.nix/files/home:/bin/sh" >> ${instDir}/etc/passwd
+  fi
+
+  if [ ! -e ${instDir}/etc/group ]; then
```
t184256: That's a really good idea.

src/pkgs/files/login.nix
t184256: If you want to get rid of the Android-given username that much, you should probably also remove passing it to the .login-inner below.

Gerschtli:
Okay, my idea was as follows:

We build the initial zip ball, as plain and basic as possible. On first boot, we want to replace the static links and fixed paths in the login script to be maintainable after a nix-collect-garbage. With the requirement to not force the user to set up and use home-manager, I implemented the `basic-environment` env. This derivation ensures that all necessary scripts and packages are installed. Of course there could be more parameterized and the final home-manager module will allow much more fine grained control over the dependencies and scripts. But for now (for the PR) I thought it would be enough to just install the basic-environment.

I needed to come up with a solution to provide a reasonable working environment for both use cases: with and without home-manager.

t184256:
> I implemented the basic-environment env. This derivation ensures that all necessary scripts and packages are installed. Of course there could be more parameterized and the final home-manager module will allow much more fine grained control over the dependencies and scripts. But for now (for the PR) I thought it would be enough to just install the basic-environment.

In general, I'm fine with merging any small improvements that work and redoing stuff later if I see fit.

But, since we've started disscussing it anyway, let's pause for a moment. I see some ways to parametrize basic-environment if it's pulled in via h-m (function parameters, h-m module). But how would one parametrize it if it's installed through nix-env? Is there some way that I don't know of?

Gerschtli:
Of course! I could not find any better ones:

The packages `pinnedPkgs`, `crossPkgs` and `crossStaticPkgs` are always pinned.

On `initialBuild` (when building the zip ball) `buildPkgs` is equal to `pinnedPkgs` to ensure complete reproducabilty of the zip ball.

When using this repo as nix-channel, `initialBuild` is false. This means that `buildPkgs` is the current installed version of nixpkgs. This allows the user to update scripts to use the newest dependencies.

I'm open for better names of these attributes!

src/pkgs/bootstrap.nix
```diff
@@ -26,8 +26,6 @@ buildPkgs.runCommand "bootstrap" { } ''
   cp ${files.nixConf} $out/etc/nix/nix.conf
   cp ${files.resolvConf} $out/etc/resolv.conf
 
-  touch $out/etc/UNINTIALISED
```
t184256: Wow. =D Yeah, who needs marker files when one can just write an expression that builds a script that will then later call another nix to regenerate itself. And it makes perfect sense.

Gerschtli:
> But how would one parametrize it if it's installed through nix-env? Is there some way that I don't know of?

I think there was a way via configuring the parameters in `~/.config/nixpkgs/config.nix` but I need to find the exact way to do it..

t184256:
> The packages pinnedPkgs, crossPkgs and crossStaticPkgs are always pinned.

Oh, my bad, I see now. Concerns waived.

src/pkgs/files/login-inner.nix
```diff
@@ -1,14 +1,14 @@
...
+{ initialBuild, instDir, packageInfo, writeTextDir }:
 
-writeText "login-inner" ''
+writeTextDir "bin/login-inner" ''
   set -e
 
   [ "$#" -gt 1 ] || echo "Welcome to Nix-on-Droid!"
 
-  [ "$#" -gt 1 ] || echo "If nothing works, use the rescue shell and read ${instDir}/bin/.login-inner"
+  [ "$#" -gt 1 ] || echo "If nothing works, use the rescue shell and read ${instDir}/bin/login-inner"
```
t184256: Why do you rename `.login-inner`? The previous rationale for hiding it was that I couldn't come up with a scenario where it wasn't used without `login`.

Gerschtli:
Given a nix expression:
```nix
with import <nixpkgs> { };

writeText "example" ''
  ${builtins.toJSON config.example}
''
```

And a `~/.config/nixpkgs/config.nix` with:
```nix
{
  example = {
    value1 = 5;
    value2 = "test";
  };
}
```

The result file would contain:
```json
{"value1":5,"value2":"test"}
```

Through this way, we could parameterize the derivation installed via `nix-env`.

t184256:
Neat. BTW, is there a way to configure that `config` from within h-m?

src/pkgs/files/login-inner.nix
Gerschtli: The main reason was to silence `writeTextDir` because it generates the derivation name from the file name. But derivation names must not start with a dot..  Furthermore `login-inner` is no executable anymore.

To be honest, it was a quick fix, but I don't know if `login-inner` really belongs into `/bin`.

t184256: TBH, it does not. You can put it somewhere else, sounds right.

Gerschtli:
One could generate a `config.nix` via home-manager but that seems like a bad idea. The better way to do it would be to `.override` the derivation in home-manager and fallback to the `config` attribute if no override got applied.

src/pkgs/files/login-inner.nix
Gerschtli: Do have a fitting path in mind?

src/pkgs/files/login.nix
Gerschtli: Why do you think it would be needed to remove passing the username to login-inner? That way I would need to hardcode the name twice and login-inner gets called only from login script.

src/pkgs/files/login-inner.nix
t184256: Not really. I'd probably go with something like `.../usr/helpers/.login-inner`, but I have no strong preference.

src/pkgs/files/login.nix
t184256: If `/etc/passwd`/`/etc/group` generation has to stay here, than yes, you'll need to hardcode it twice.
If it could be moved into the initial setup, then there would be no need to have it exported in the outer login at all.

Now the question is what made me do that in the outer login back then.

Gerschtli:
It seems to be tricky to get the current UID in nix.

Furthermore `proot` does not really want to get built on the phone. Need to further investigate what is holding it back..

t184256:
> It seems to be tricky to get the current UID in nix.

I'm trying to see if it could be deferred to the point where basic-environment is already present.

> Furthermore proot does not really want to get built on the phone. Need to further investigate what is holding it back..

Yeah, this one just couldn't be simple.

Gerschtli:
To test this PR, you can use https://downloads.tobias-happ.de/bootstrap as bootstrap url. I changed `login-inner` to use https://downloads.tobias-happ.de/bootstrap/channel.tar.bz2 as nix-channel which is just a tar ball of the repo.

t184256:
> I changed login-inner to use https://downloads.tobias-happ.de/bootstrap/channel.tar.bz2 as nix-channel which is just a tar ball of the repo.

Thanks for the hint, I was just wondering what's wrong with my build.

t184256:
Compiling statically-linked proot on-device invokes compiling the whole toolchain, which takes about an eternity or so. What's even worse, this will happen quite a lot on nixpkgs updates. If we land this without finding some way to avoid perpetually recompiling the toolchain on each update, I'll really need help from some resourceful aarch64 folks to combat it. Sigh.

Are you, by chance, going to attend NixCon 2019?

t184256:
OK, static compilation was easy (though it still pulls gcc and slooowly compiles talloc/proot on first boot, but that is way better than the toolchain).

Another problem: the symlinks in bin/ are indirect, meaning that they do not resolve correctly outside of proot env. I've thought about it, and I think that the best way to fix that is actually from the Java-land. Currently working on that.

Gerschtli:
Ah, nice idea to retrieve the uid and gid!

Gerschtli:
> Are you, by chance, going to attend NixCon 2019?

Sadly not, maybe next year :)

Gerschtli:
I think we should either just use the proot binary shipped via the zip ball or provide a cached version via cachix. To force the rebuild on the device is not a working solution for me..

t184256:
I've just rechecked and it compiles fine for me with `nix build --show-trace -f src --argstr arch aarch64 --argstr nixOnDroidChannelUrl https://github.com/t184256/nix-on-droid-bootstrap/archive/testing.tar.gz`, and my `testing` is pointing to e7c2d5298b960b31b47caa66f83afe0d509cb576. Relogin does not work though because of broken symlinks, so it's not tested to work yet.

What do you mean by "not a working solution"? takes too much time? space? doesn't compile? doesn't work afterwards?

Gerschtli:
What I mean is that my phone is already compiling 10 minutes. I don't know how long it will take, but if a fresh user tries nix-on-droid, he will surely be disappointed to wait an unknown amount of time before being able to do anything useful with the app..

t184256:
True.

But the prospect of managing everything with Nix sounds so good.

And I can see two ways of alleviating that:

1) I can periodically push to cachix.

2) (if we roll high enough on diplomacy) we can add proot-static to nixkpgs itself, so that the Hydra itself will rebuild it and push it to the official cache.

Gerschtli:
How do you want to build proot via Hydra? You need an android device as underlying architecture if I'm not missing something.

We could use the pinned packages to build proot and update the pinned version regularly. In this way, there would not be any chance that the user would need to rebuild it locally.

t184256:
You don't. You just need a real aarch64 device, and Hydra has them. Same logic as with all the other packages.

(provided that whatever is built now will actually work and is not some utter garbage)

> We could use the pinned packages to build proot and update the pinned version regularly. In this way, there would not be any chance that the user would need to rebuild it locally.

Hey, yeah, that's also an option.

Gerschtli:
Ah that sounds good.

I think, we should pin the nixpkgs and push the proot/talloc builds for these special versions to completely avoid rebuilds on the android phones.

Meanwhile there should me an aarch64 build device for pushing the proot/talloc builds to cachix. (I would appreciate it if you would push the builds to cachix so I can test this PR locally :D)

Gerschtli:
BTW, I would love to refactor this mess we made :D I'm already confused with all the naming stuff and what parameter `callPackage` will inject in each file. I don't know how you feel, but this hurts :D It is definitely alright as a proof of concept, but it is getting more and more complex..

But to start refactor, we should finish the proot/talloc thing first. Otherwise this will result in big merge conflicts..

t184256:
> I think, we should pin the nixpkgs and push the proot/talloc builds for these special versions to completely avoid rebuilds on the android phones.

Sounds sensible.

> Meanwhile there should me an aarch64 build device for pushing the proot/talloc builds to cachix.

Will work on that after I finish fighting Android Studio and solve the broken symlinks problem.

I now have access to two devices, which will also enable me to test both wipe-reinstall and simply-rebuild scenarios =)

> (I would appreciate it if you would push the builds to cachix so I can test this PR locally :D)

 It should be just ~20 minutes of compiling.

> But to start refactor, we should finish the proot/talloc thing first.

Not only. To land this, we need

1. the symlinks problem solved
2. proot/talloc pushed to cachix
3. proot/talloc pinned
4. some basic smoke testing to check that it's somewhat functional

and I'm not as fast as you are =) You can do 3 while I'm fighting the first two.

t184256:
> (provided that whatever is built now will actually work and is not some utter garbage)

uhh... the output is dynamically-linked. gotta fix that too

t184256:
I'm giving up for now.

If I link proot with glibc, it fails with `Bus error`. Building a usable proot (i.e., linking it to bionic) on-device is too hard for me now. It is doable (termux can do that and that's how I've started in the first place), but it's complicated. I'll try to remove proot-managing commits from the branch.

t184256:
OK, now it's your turn to review =) I've done some testing and fixed what I've found (mostly h-m installation being broken).

I've removed some of the commits regarding native proot building, so it'll be a force-push. The originals are present in a `FAILED-on-device-proot-building`branch  if you need them.

Testing requires a new version of the app, or the sessions after the first one would fail. Channel URL can be specified as a build parameter.

default.nix
```diff
@@ -0,0 +1,38 @@
+# Licensed under GNU Lesser General Public License v3 or later, see COPYING.
+# Copyright (c) 2019 Alexander Sosedkin and other contributors, see AUTHORS.
+
+{ }:
```
Gerschtli: Can you explain the reason behind this change?

t184256: In h-m config, we import it as a function, so I made it into one.

src/pkgs/files/login.nix
```diff
@@ -1,29 +1,85 @@
...
   #!/system/bin/sh
   set -e
 
-  export USER=$(/system/bin/whoami)
+  ORIGINAL_PATH="$PATH"
+  ANDROID_LIKELY_PATH="/sbin:/system/sbin:/system/bin:/system/xbin"
+  ANDROID_LIKELY_PATH="$ANDROID_LIKELY_PATH:/vendor/bin:/vendor/xbin"
+  PATH="$ANDROID_LIKELY_PATH:$OLD_PATH"
+
+  resolve_link() {
```
Gerschtli: Oh boy, this is really complicated.. Can you explain in words what exactly was the problem and we do you need so many cases?

AFAICT the symlinks of login (failing in the android app) and proot-static, sh and login-inner are the problem right? So the symlink blob in the `~/.nix-profile` is the real source of problems, I see.. I need a moment to work this through :D

src/pkgs/files/hm-install.nix
```diff
@@ -0,0 +1,33 @@
...
+
+  export NIX_PATH=$HOME/.nix-defexpr/channels''${NIX_PATH:+:}$NIX_PATH
+
+  if [ ! -e $HOME/.config/nixpkgs/home.nix ]; then
+    echo "Creating an initial home-manager configuration in ~/.config/nixpkgs/home.nix ..."
+    ${coreutils}/bin/mkdir -p $HOME/.config/nixpkgs
+    ${coreutils}/bin/cp -n /etc/home.nix.default $HOME/.config/nixpkgs/home.nix
+  fi
+
+  ${nix}/bin/nix-env --set-flag priority 120 basic-environment
```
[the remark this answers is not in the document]
t184256: Yeah, I thought about it a lot, and I think it's not bad.

How it was: we uninstall basic-environment, `/etc/resolv.conf` disappears, home-manager installation reliably fails, the whole env is broken.

Option 1: we lower the priority of basic-environment, we install the same thing through h-m, we remove the original ones. This works, but hey,

Option 2: not removing it works too. This way we don't have to worry about the user removing it through h-m, and uninstalling h-m won't break the env too. If parts of it get updated/overridden, they will shadow the old ones.

Upsides: harder to break the resulting environment.

Downsides: if you really want to uninstall something from it, you'll have to remove it from both. But then again, that'd be quite disastrous. Old copy is around and not garbage-collected.

src/pkgs/files/login.nix
t184256: Ha-ha, it totally is.

Short explanation: the real files are hidden under layers and layers of symlinks. This would not be a problem at all if only those were working symlinks. They are not.

Complication 1: symlinks can point to, for example, `/nix/store/123-whatever`. This doesn't make sense for the outside Android system. You see such a symlink (not pointing to `/data/data/com.termux.nix/files`), you prepend `/data/data/com.termux.nix/files/usr` to it. Now it works.

Complication 2: symlinks can eventually point to `/long/path/that/has/a/broken_symlink/and/something/after/it`. When you have this thing, you have to iterate over it and see which one of the parents is a broken symlink, fix it, resolve it, add the remainder and continue resolving. It's bonkers.

Here `login` finds `proot-static`. There's also a Java-land implementation of this that finds `login`: https://github.com/t184256/nix-on-droid-app/commit/f4d6e55cda63735a504669fbdf0c1fdd50cbe657

You can marvel at my notes on how `login` is found:

```
    // /d/d/c.t.n/f/usr/bin/login -> /d/d/c.t.n/f/usr/nix/v/n/p/p/USERNAME/profile/login
    // /d/d/c.t.n/f/usr/nix/v/n/p/p/USERNAME/profile/login is broken
    // /d/d/c.t.n/f/usr/nix/v/n/p/p/USERNAME/profile -> profile-1-link
    // /d/d/c.t.n/f/usr/nix/v/n/p/p/USERNAME/profile-1-link -> /nix/store/...-user-environment
    // /nix/store/...-user-environment/bin
    // /d/d/c.t.n/f/usr/nix/store/...-user-environment/bin -> /nix/store/...-basic-environment/bin
    // /d/d/c.t.n/f/usr/nix/store/...-basic-environment/bin/login -> /nix/store/...-login/bin/login
```

or uncomment the debugging output in the shell script and see a whole screen of detective work to find `proot-static`.

Also, shipping `login` that is a broken symlink broke my tablet setup, actually, as I have an external process (Easer) that gains root, enters the Nix environment and executes an evdev/uinput keyboard remapper when I connect my Bluetooth keyboard. So, maybe one day I'll move this symlink resolver into a separate file to reuse it in other 'Android chroots into Nix' scenarios.

But hey, now we have an upgrade path that doesn't require wiping. Worth it.

src/pkgs/files/login-inner.nix
```diff
@@ -1,28 +1,56 @@
...
+      echo "itself."
+      echo
+      echo "You can go for the bare Nix setup or you can configure your phone via"
+      echo "home-manager. For that simply run hm-install."
+      echo
+    '' else ""
+  }
+
+  [ "$#" -gt 0 ] || echo "Sourcing Nix environment..."
+  . $HOME/.nix-profile/etc/profile.d/nix.sh
```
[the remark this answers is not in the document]
t184256: It's a different `nix.sh`.

src/pkgs/files/hm-install.nix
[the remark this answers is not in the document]
Gerschtli: That seems to be a good idea. We should at least write some documentation after this PR is finished to explain what we are doing here to be transparent to the user.

src/pkgs/files/hm-install.nix
```diff
@@ -0,0 +1,33 @@
...
+
+  echo "Installing home-manager..."
+  ${nix}/bin/nix run nixpkgs.nix -c ${nix}/bin/nix-shell '<home-manager>' -A install
+
+  echo "home-manager is installed. Please, restart the session."
```
Gerschtli: Why not `. "$HOME/.nix-profile/etc/profile.d/hm-session-vars.sh"` instead of this notice?

t184256: Because it won't propagate to the parent shell.

src/pkgs/files/hm-install.nix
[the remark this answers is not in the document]
t184256: I've added a comment to the default h-m config.

t184256:
LGTM. And this one is big enough already.

Will merge when you'll say it looks OK, but don't hesitate to voice more concerns (or reraise some that I forgot about, if there are any). I'm fine with minor commit rephrasing/reordering/fusing, otherwise I'll just merge them all as is, w/o squashing.

Gerschtli:
LGTM too. I will be busy the next days let's merge it as it is right now and do the next things in new PRs.

Don't forget to update the default nix-channel url to master branch. And I think we should update the README according to the newest changes.

I installed it on my phone and the initial boot and `hm-install` work without problems.

t184256:
Fine. Thanks a lot for your hard work.

I've updated the README. I've also removed the license header from the default config template, as I'm 99+% sure you didn't intend to enforce GPL on user config.

Gerschtli:
Yeah, you are right, thank you!

PR #21, opened 2019-11-03
Big Refactoring

Gerschtli:
Hey, it's me again :)

As I mentioned some time ago, there is the refactoring and cleanup PR.

Benefits of this refactoring:
- argument list is now comprehensibly as there is now only one nixpkgs version flying around
- common config and lib functions got abstracted away
- clearer intent when to use pinned or unpinned packages
- allows better configurability for the future through config attr set

Let me know, what you think of this (opinionated) refactoring.

It is not tested yet (long build times, you know), but it should work as before :)


Gerschtli:
Zip ball generation and installation on device works like a charm. Same for `hm-install`, so I'd say it's ready for review :)

src/lib/default.nix
```diff
@@ -0,0 +1,40 @@
...
+
+  userName = "nix-on-droid";
+in
+
+{
+  buildConfig = { arch, initialBuild, nixOnDroidChannelURL }: (
+    {
+      channel = {
+        nix-on-droid = nixOnDroidChannelURL;
+        nixpkgs = "https://nixos.org/channels/nixos-19.09";
```
t184256: I'm confused. Why does the default channel URL live here now? It only takes place on initial execution (and after wipes), so I think it is sensible to configure it at zipball build time, but not during normal usage.

Gerschtli:
It is not possible yet to configure anything from the user config, this is just not implemented yet.

Issue #6, opened 2019-03-13
Building proot is not automated


t184256:
Unfortunately my attempts for naively cross-compiling static-linked proot as `pkgsCross.armv7a-android-prebuilt.pkgsStatic.proot` or `pkgsCross.aarch64-multiplatform.pkgsStatic.proot` failed: the produced binaries crashed with the dreaded bus error that I don't know how to debug, so I resorted to building proot manually for now.

This may be a desperate shot in the dark, but when I type 'cross-compiling static-linked... with nix`, there's a name that rings inside my head and this name is @matthewbauer. Matthew, if you have a minute, do you have any insights on why could that produce a binary that crashes with 'Bus error'? Have you tried combining crossSystem and pkgsStatic?

ln-nl:
(it may be useful: https://github.com/NixOS/nixpkgs/pull/58504/commits/52566c3b7b5f96d198dd6e57e2db37012fd9b71b)

t184256:
Mindlessly tried building

    pkgsCross.aarch64-multiplatform.pkgsStatic.proot
    .overrideAttrs (oa: {
      version = "2019-03-12";
      src = pkgs.fetchFromGitHub {
        repo = "proot";
        owner = "termux";
        rev = "09568e21ff9fee7a6907be34a2d0fde3bacca7eb";
        sha256 = "1091si4i07349y7x7sx1r904gg3c9a3m39xkv24fakjlirpi3lpy";
      };
      makeFlags = [ "-Csrc CFLAGS=-D__ANDROID__" ];
    }))
    .override {
      enablePython = false;
    )

from that checkout, got a `proot` that dies with a Bus Error on a very early `mkdir`.

matthewbauer:
Hmm... it's possible that the proot fork only works with bionic? That gets trickier because we don't have a good way to cross execute android linker. Something like this should work:

```
let
  pkgs = import <nixpkgs> {
    crossSystem = (import <nixpkgs/lib>).systems.examples.aarch64-multiplatform-musl;
    crossOverlays = [ (import <nixpkgs/pkgs/top-level/static.nix>) ];
  };
  proot = pkgs.proot.override { enablePython = false; };
in pkgs.runCommand "test" {} ''
  ${pkgs.stdenv.hostPlatform.emulator pkgs.buildPackages} ${proot}/bin/proot --version
''
```

but there's obviously some things specific to android that could be causing the bus error.

t184256:
[...]

PR #28, opened 2019-11-24
Module system

Gerschtli:
Hey,

as this is really much to review, I will open it while I'm working on some more things before this PR is ready to merge.

I reused some concepts of nixpkgs/nixos and home-manager. I'm not that into this licence thing but I hope I marked it accordingly..

Things left to do:
- [x] First execution of `nix-on-droid switch` in `login-inner` results in closing the app. I don't have any idea why this is happening, because if I uncomment it there and run it manually, it works just fine..
- [x] Try to be mostly backwards compatible as it can get.
- [x] Add some more commands for `nix-on-droid`.
- [x] Update/add documentation
- [x] Optional: Update pinned nixpkgs version
- [x] Optional but somehow necessary in the long run: Be able to update proot-static (as download from https://nix-on-droid.unboiled.info/proot-static for example)

Resolves https://github.com/t184256/nix-on-droid-bootstrap/issues/26

Sorry for the big commits, it is really hard to split the changes in small little chunks..


t184256:
Not a real review yet: I do remember about this one, I have skimmed through it once, I've spent some time trying to clean-install this and the app keeps closing around the end of the installation. Does clean-install work for you?

Gerschtli:
This is the very first point in my todo list. I can push a somehow working version by tomorrow, but I spent the last few days figuring out, why the process is exiting on the initial build. It's so strange, I tried everything but can't find the cause. It is exiting the proot session right after the nix-on-droid switch, sometimes it prints out the following echo's but sometimes it stops right after the switch. I don't have any idea on how to solve this..

Gerschtli:
To clarify my current problems:

On the first boot the `modules/environment/login/login-inner.nix` file exits right after `nix-on-droid switch`. Not even a single `echo` is printed out after that. I have no idea, why this is happening. The workaround is to call `/bin/login` again (see `modules/environment/login/login.nix`). Do you have an idea why this is happening?

The other thing I ran into was that `nix-env -i` installed the first profile in `/nix/var/nix/profile/default` and therefore did not create the link in `~/.nix-profile`. I worked around this problem through setting this link manually in `modules/environment/login/login-inner.nix` line 25. According to the source code I looked into this could be caused by `getuid()` returning 0 instead of the correct UID but `id -u` returns the right UID.

Gerschtli:
Fixed the initial build and removed the workaround. It's ready to review now.

The missing pieces are documentation updates and backwards compatibility concerns. For the bc we can either go for a breaking change or provide the basic-environment with `nix-on-droid` added. The latter means maintaining both setups for a while. The choice is up to you.

For the future we should figure out on how to render a nice view for all possible options, I would be okay to just add some information in the README and redirect to reading the source for now. What do you think?

Gerschtli:
And Travis is failing because the build outputs generated by `ci.nix` are not available through `nix-on-droid.cachix.org`.

modules/environment/login/default.nix
```diff
@@ -0,0 +1,94 @@
...
+          $DRY_RUN_CMD chmod $VERBOSE_ARG u+w /bin/.proot-static.tmp
+          $DRY_RUN_CMD mv $VERBOSE_ARG /bin/.proot-static.tmp /bin/.proot-static.new
+        fi
+      '';
+    };
+
+    environment.files = {
+      inherit login loginInner;
+
+      prootStatic =
```
t184256: `path '/nix/store/bfxybmnxc39fjid3jsxsdfxnnszyijh0-proot-termux-unstable-2019-09-05-aarch64-unknown-linux-android' does not exist and cannot be created`

Gerschtli: Yes, you need to build and push proot into cachix as it gets fetched via its store path.

t184256: I'm in progress, but that's something awkward.

t184256: Built and pushed.

The reason is that it should point at a cross-compiled version, and hardcoding it is the only way to have the correct hash on-device?

There should be some way around that, otherwise that's pretty limiting: complicates forking to no end, updates become undeliverable without cachix...

Gerschtli: Either that way or uploading proot after each nixpkgs update to some cdn. We need a way to get proot-static updated on the device..

Gerschtli: The problem is, that currently we only push the build packages to cachix.

We need a build chain with aarch64 and i686 machines that build and push proot (and possible patched packages via overlays) to cachix, then we can use a nix expression instead of a hardcoded store path.

modules/environment/login/login.nix
```diff
@@ -0,0 +1,37 @@
...
+
+writeScript "login" ''
+  #!/system/bin/sh
+  set -e
+
+  export USER="${config.user.userName}"
+  export PROOT_TMP_DIR=${installationDir}/tmp
+  export PROOT_L2S_DIR=${installationDir}/.l2s
+
+  if test -e ${installationDir}/bin/.proot-static.new; then
```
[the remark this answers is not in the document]
Gerschtli: In the activation script, I am installing `proot-static`, `login` and `login-inner`. My first attempt was to replace these files directly on activation. When trying to replace `proot-static`, I got a "file is busy" error message. When trying to replace `login-inner`, the execution of `login-inner` stopped immediately (after finishing the started child processes like in this case `nix-on-droid switch`). Replacing `login` works without problems.

To clarify: These problems occur everytime I install a new version of any of these files, not just on the initial boot.

The installation of the new files need to be at login time, because there I can safely replace the old files.

As I write this, I think I forgot the case when there are multiple running sessions. I need to somehow include a check, if there is no other session running and only then replace these files..

modules/home-manager.nix
```diff
@@ -0,0 +1,59 @@
...
+  options = {
+
+    home-manager = {
+      config = mkOption {
+        type = types.nullOr hmModule;
+        default = null;
+        description = "Home Manager configuration.";
+      };
+
+      useUserPackages = mkEnableOption ''
```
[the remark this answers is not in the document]
Gerschtli: There are two user facing differences:
* In `nix-env -q` you can see that there is `nix-on-droid-path` and `home-manager-path` installed. These packages are simple env's with all packages to install. With enabling this option there would only be one (our) package to be installed.
* If you install the same package via `nix-on-droid` and `home-manager` or packages with files with same name, this could lead to name clashes or maybe `nix-env -i` could throw errors trying to install both env's.

It's a nice feature and I see no disadvantages of it. Maybe we should enable it by default?

modules/default.nix
```diff
@@ -0,0 +1,37 @@
...
+    modules = [
+      {
+        _module.args = { inherit pkgs; };
+      }
+    ]
+    ++ optional (!initialBuild && hasConfigFile) configFile
+    ++ optional (!initialBuild && !hasConfigFile && pkgs.config ? nix-on-droid) pkgs.config.nix-on-droid
+    ++ optional initialBuild config
+    ++ import ./module-list.nix;
+  };
```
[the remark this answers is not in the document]
Gerschtli: That would be possible if you set the following in your `nix-on-droid.nix`.
```nix
{
  home-manager.config = import ./home.nix;
}
```

How about adding this line to the config in a migration script if `home.nix` is present? I don't feel good about letting `nix-on-droid` automagically find a `home.nix` and source it. But of course we could wrap the line in a new option.

modules/build/initial-build/nix-on-droid.nix.default
```diff
@@ -0,0 +1,38 @@
...
+  # After installing home-manager channel like
+  #   nix-channel --add https://github.com/rycee/home-manager/archive/master.tar.gz home-manager
+  #   nix-channel --update
+  # you can configure home-manager in here like
+  #home-manager.config =
+  #  { pkgs, ... }:
+  #  {
+  #    # insert home-manager config
+  #  };
+}
```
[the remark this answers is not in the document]
Gerschtli: Yeah, why not.

Just out of curiosity: Do you have a use case in mind for this PR or is it that it is nice feature?

modules/environment/etc/setup-etc.sh
```diff
@@ -0,0 +1,85 @@
+# Licensed under GNU Lesser General Public License v3 or later, see COPYING.
+# Copyright (c) 2019 Alexander Sosedkin and other contributors, see AUTHORS.
+
+# inspired by https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/system/etc/setup-etc.pl
+
+etc="${1}"
+static="/etc/static"
+new_etc="${2}"
+
+function atomic_symlink() {
```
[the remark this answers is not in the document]
Gerschtli: Haha, I learned so much working on this PR. I read so much source code of home-manager, nixpkgs and nix-darwin.

modules/environment/files.nix
```diff
@@ -0,0 +1,36 @@
+# Licensed under GNU Lesser General Public License v3 or later, see COPYING.
+# Copyright (c) 2019 Alexander Sosedkin and other contributors, see AUTHORS.
+
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+{
```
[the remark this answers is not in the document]
Gerschtli: Yes, because we need these files but I want these to be configurable in the future.

nix-on-droid/nix-on-droid.sh
```diff
@@ -0,0 +1,169 @@
+#!@bash@/bin/bash
```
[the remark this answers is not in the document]
Gerschtli: Yeah me too. I tried to come up with a quite minimal feature set.

pkgs/lib/load-nixpkgs.nix
```diff
@@ -0,0 +1,18 @@
...
+
+let
+  defaultNixpkgsArgs = {
+    config = { };
+    overlays = [ ];
+  };
+
+  # head of nixos-19.09 as of 2019-11-28
+  # note: when updating nixpkgs, update store paths of proot-termux in modules/environment/login/default.nix
+  pinnedPkgsSrc = builtins.fetchTarball {
```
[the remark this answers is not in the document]
Gerschtli: That way we do not need a nixpkgs instance to fetch another nixpkgs version. It is a lot cleaner to just have one nixpkgs instance around if we can simply use builtin functions of nix

Gerschtli:
I like the idea of a migration script. But we need to define what this script should do.

* Copy `nix-on-droid.nix.default` to `~/.config/nixpkgs`
* Find `home-manager` config file (either `HOME_MANAGER_CONFIG` set or `~/.config/nixpkgs/home.nix` present) and update `nix-on-droid.nix` to read this config file
* Run `nix-on-droid switch`
* Uninstall `basic-environment`

Is there anything I forgot?

Gerschtli:
And do you have an idea on how this migration script gets invoked? Ideally the should do
```
nix-channel --update nix-on-droid
nix-env -u basic-environment
```
And everything else runs automatically, but I do not know how start the script..

Gerschtli:
To migrate, the user has to update the channel and run `nix-shell '<nix-on-droid>' -A migration`.

If running `nix-env -u --always basic-environment` this command is displayed with an small explanation.

Things left to do: clean old and now broken symlinks in `/etc` and optionally find a way to show this message on `nix-env -u basic-environment` without providing `--always`.

modules/environment/login/login.nix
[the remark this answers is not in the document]
Gerschtli: If we do not need to support the old setup the broken symlinks logic can be removed from the app.

BTW, the problem with multiple sessions has been fixed

modules/home-manager.nix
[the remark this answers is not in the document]
Gerschtli: For consistent behaviour I would prefer disabling this option like in nixos and nix-darwin modules.

modules/default.nix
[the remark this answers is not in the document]
Gerschtli: The migration scripts adds either
```nix
{
  home-manager.config = import ./home.nix;
  home-manager.useUserPackages = true;
}
```
or
```nix
{
  home-manager.config = import (builtins.getEnv "HOME_MANAGER_CONFIG");
  home-manager.useUserPackages = true;
}
```
depending on the present setup.

Gerschtli:
The last thing left (besides your incoming feedback of course ;) ) is the documentation.. I really don't know what to document and in which way. Maybe if you are reading through this PR, you can pass me some pointer. For me it is hard to document a thing because now it seems all so trivial to me :D

Gerschtli:
The testing is a bit tricky, I hope I found every bug by now :)

modules/environment/login/default.nix
Gerschtli: For the build chains for aarch64 (https://docs.travis-ci.com/user/multi-cpu-architectures/) and i686 (https://unix.stackexchange.com/questions/292779/how-to-build-32-bit-application-on-64-bit-system-using-nix-package-manager/292829), there could be ways to achieve this via travis as build tool (as long as the build times are reasonable, otherwise we need to another solution, maybe get support by hydra.nixos.org?).

Gerschtli: But I'm not sure, how this works together with the fact, that the destination arch is an android phone..

t184256: (something I can answer quickly) I'm fairly sure that it has to be build with a bionic toolchain (for the reasons not entirely clear to me), and that means x86_64 buildhost.

Gerschtli: Can you explain, what you mean by bionic buildchain?

t184256: The one that is used to compile libraries for Android apps. Clang and bionic bundled by Google and distributed under the name of NDK. If proot is not compiled with/against it, it mysteriously terminates: #6.

Gerschtli: Ah I see, and compiling it with bionic does have a different nix store path than compliling it natively, right?

t184256: Correct.

t184256:
I've pushed both hardcoded proot derivations to my cachix and it still attempts to install NDK on-device... What am I doing wrong?

Gerschtli:
There seems to a dependency between proot and ndk-bundle, it is the case for me too.

Gerschtli:
Hey, do you know when you will be able to review this? Or do you already have some pointers on what to improve?

For the hardcoded proot path: I really don't have any idea on how to make this more robust.. Another way I can think of, would be to upload the proot binary to your cdn and write a derivation which just dowloads it from there..

t184256:
OK, I have a bit of time, let's reignite this.

Two things that bug me the most are

1) proot/talloc derivations having 2.4 gigs of dependencies that totally shouldn't be pulled onto the devices. And they shouldn't have any, actually, these are really self-contained. ([talloc deps](https://gist.github.com/t184256/702242d2bce80024d54c9aa9b45c2718)). This is definitely a blocker.

(2.4G is taken from `nix copy --no-check-sigs --to local?root=SOME-ABSOLUTE-PATH /nix/store/bfxybmnxc39fjid3jsxsdfxnnszyijh0-proot-termux-unstable-2019-09-05-aarch64-unknown-linux-android`)

2) the hardcoded proot path is actually an OK thing if you are me. If one ain't me, it's a bit of a problem.
  If someone wants to fork nix-on-droid to test something real quick and forget it, then it's avoidable by baking initial proot/talloc into the bootstrap zipball. If one wants to maintain it for a while, then they'll have to create own cachix, push to it, subscribe users to it and only then it'll work. That's not that nice, but it's not that bad either, we can just document it.

So, could you please somehow cut off the proot derivation dependencies and bake it into the zipball?

Gerschtli:
Sadly, I don't have much knowledge about this static building and so-libraries. The only thing I was able to find out was that the path `/nix/store/k7skwbz9yzmiy7wj8vjkfc5d3djk64bc-bionic-prebuilt/lib` occurs in `libtalloc.so.2.1.14`. Do you have some pointer, what I could try to remove this path or at least where this path got injected?

This seems to be the only reference from proot/talloc to the ndk bundle.

t184256:
Oh man, that's it! It's not present in `proot` and `talloc.a`, but you're right, it's there in `talloc.so`, and we include it. Then nix greps for `/nix/store` and adds the NDK dependency we do not want at all. Wow.

I've pushed a straightforward fix, and I can already say it has shrunk the installation size from 3.17G to 486M.

Gerschtli:
Ah, that's awesome!

Then the last step should be documentation. I will update the readme regarding basic build of zip ball and basic usage.

My approach would be to open separate issues for providing a nice and generated visualisation of all possible options and build proot on device to replace the hardcoded paths.

Everthing else should be considered ready to merge! :)

t184256:
One more thing is bundling/injecting proot closure into the zipball. I'm currently stuck at being unable to call `nix` or `nix-store` from the build script.

Gerschtli:
This does not look that trivial.. I tried using `nix-store --dump` in the `nix-directory`  derivation:
```bash
${nix}/bin/nix-store --dump ${tallocStatic} > talloc.dump
${nix}/bin/nix-store --dump ${prootTermux} > proot.dump
USER=nix-on-droid ${prootCommand} "$PKG_NIX/bin/nix-store" --restore ${tallocStatic} < talloc.dump
USER=nix-on-droid ${prootCommand} "$PKG_NIX/bin/nix-store" --restore ${prootTermux} < proot.dump
```

The store paths are present in the zip ball but the derivations will still be downloaded.. Besides that: Is it worth it? The initial `nix-on-droid` build downloads ~100MB which in comparison to the nixpkgs channel is nothing. But thats just my 2 cents.

t184256:
Yeah, dump-restore doesn't register it in the sqlite db, while export just doesn't work altogether.

My idea was that this way you don't need cachix for the initial installation => it's way easier for someone to fork the repo, change something and just try out some stuff.

I'm kinda tired of trying to pull that of from inside the nix sandbox and I'll probably play dirty and just write an external script for that =/

t184256:
I've pushed a shell script that does what I want to happen: autochange proot hashes, inject their closures to the zipfile. Not sure if any of these is possible with nix magic, maybe you have any ideas.

Will test during December / early January.

build.sh
```diff
@@ -0,0 +1,58 @@
...
+	cp ../proot-$arch.closure ./proot.closure
+	zip --quiet -r ../bootstrap-$arch.zip .
+	popd >/dev/null
+	chmod -R +w out/repack-$arch
+	rm -rf out/repack-$arch
+	echo $arch: done
+done
+
+
+if [ -z "$rsync_to" ]; then
```
Gerschtli: How do you set this value?

modules/environment/login/login-inner.nix
```diff
@@ -0,0 +1,93 @@
...
+
+      if [ -r ${config.build.installationDir}/proot.closure ]; then
+        echo "Injecting bundled talloc closure..."
+        ${nix}/bin/nix-store --import < ${config.build.installationDir}/talloc.closure
+        ${coreutils}/bin/rm -f ${config.build.installationDir}/talloc.closure
+        echo "Injecting bundled proot closure..."
+        ${nix}/bin/nix-store --import < ${config.build.installationDir}/proot.closure
+        ${coreutils}/bin/rm -f ${config.build.installationDir}/proot.closure
+      else
+        echo "No bundled talloc/proot closures found, which is OK for official builds..."
```
Gerschtli: For debugging purposes this may be helpful, but I prefer not to print this to a normal user

build.sh
```diff
@@ -0,0 +1,58 @@
...
+	grep "$arch = \"/nix/store/" modules/environment/login/default.nix
+
+	echo $arch: building nix-on-droid...
+	nix build --show-trace -f pkgs --argstr arch $arch --argstr nixOnDroidChannelURL $nixOnDroidChannelURL bootstrapZip -o out/nix-on-droid-$arch
+
+	echo $arch: injecting talloc/proot for initial bootstrap...
+
+	cat out/nix-on-droid-$arch/bootstrap-$arch.zip > out/bootstrap-$arch.zip
+	nix-store --export --readonly-mode $talloc > out/talloc-$arch.closure
+	nix-store --export --readonly-mode $proot > out/proot-$arch.closure
```
Gerschtli: Why not getting the complete closure? This way you do not need to export talloc manually
```suggestion
	nix-store --export --readonly-mode $(nix-store -qR $proot) > out/proot-$arch.closure
```

Gerschtli:
Didn't come up with a better solution because `/nix/var` of the host machine is not available during build time. And as this is only necessary if you fork this repo and change the way proot/talloc is built, it seems to an acceptable solution if we document it accordingly.

build.sh
t184256: Externally, to an address of the server/directory behing https://nix-on-droid.unboiled.info.

modules/environment/login/login-inner.nix
t184256: I concede, this isn't particularly helpful.

build.sh
t184256: Because I didn't know that trick!

Gerschtli:
I thought maybe we could provide some utility scripts for developers:

- Build and upload zip ball for one arch
- Pack and upload a custom channel
- And a script which does both

This would make contributions and testing a lot easier for new developers.

The `build.sh` is a good starting point, maybe adding command line arguments.

For example my test script was like this:
```bash
nix build -f ./pkgs --argstr arch aarch64 --argstr nixOnDroidChannelURL https://downloads.tobias-happ.de/bootstrap/channel.tar.bz2 bootstrapZip
tar --exclude=".git" --exclude="channel.tar.bz2" --exclude="result" --transform 's,^./,nix-on-droid/,' -cjf channel.tar.bz2 .
scp channel.tar.bz2 result/bootstrap-aarch64.zip <path-to-server>/bootstrap
```

But for the sake of this PR, I should open a separate issue/PR for that.

Gerschtli:
I managed to replace the `build.sh` with a pure nix solution :)

t184256:
> I managed to replace <your bash spaghetti of the day> with a pure nix solution

I'm starting to see a trend here =D I'm all for it. Looks fragile and internals-heavy, but we can tweak it when we bump nixpkgs and it breaks.

Gerschtli:
Haha :D I just remembered that nixos iso images are also generated via nixpkgs so I searched a bit. BTW looks stable enough to be as it is used in several places in nixpkgs.

t184256:
Trivial migrations of working setups went OK; upgrading my 'stable system' triggered one more problem due to my setup being old enough to lack basic-environment in non-h-m profile. I've decided to address that through tweaking the migration message.

I'm done with this pull request. Our users should be few and tech-savvy and there are more precautions already in place than it's necessary. 

This was monumental, again. Thank you, and hope I'll see less monumental PRs (until Android 10 forces us to bundle proot with APK or something).

I'm merging this, will fix stuff as it breaks.

Gerschtli:
Wow, thank you! You can look forward to more smaller PRs as the base is now there for everything we want to do :)

PR #38, opened 2020-01-05
Increase priority value to avoid conflict

Gerschtli:
When enabling `home-manager.useUserPackages`, there is a conflict of duplicate package installations. Through increasing priority value this can be avoided.


Gerschtli:
As the modified priority is temporary because home-manager-path will be uninstalled later in the same activation script, there should not be any confusion for the user.

PR #41, opened 2020-01-07
Add nixpkgs.config and overlays option

Gerschtli:
Now it is possible to define config and overlays in nix-on-droid config which is a necessary step for https://github.com/t184256/nix-on-droid-bootstrap/issues/23

### DIFF
--- a/modules/nixpkgs.nix
+++ b/modules/nixpkgs.nix
@@ -1,0 +1,170 @@
+# Copyright (c) 2019-2020, see AUTHORS. Licensed under MIT License, see LICENSE.
+
+# Inspired by
+# https://github.com/rycee/home-manager/blob/master/modules/misc/nixpkgs.nix
+# (Copyright (c) 2017-2019 Robert Helgesson and Home Manager contributors,
+#  licensed under MIT License as well)
+
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  isConfig = x:
+    builtins.isAttrs x || builtins.isFunction x;
+
+  optCall = f: x:
+    if builtins.isFunction f
+    then f x
+    else f;
+
+  mergeConfig = lhs_: rhs_:
+    let
+      lhs = optCall lhs_ { inherit pkgs; };
+      rhs = optCall rhs_ { inherit pkgs; };
+    in
+    lhs // rhs //
+    optionalAttrs (lhs ? packageOverrides) {
+      packageOverrides = pkgs:
+        optCall lhs.packageOverrides pkgs //
+        optCall (attrByPath ["packageOverrides"] ({}) rhs) pkgs;
+    } //
+    optionalAttrs (lhs ? perlPackageOverrides) {
+      perlPackageOverrides = pkgs:
+        optCall lhs.perlPackageOverrides pkgs //
+        optCall (attrByPath ["perlPackageOverrides"] ({}) rhs) pkgs;
+    };
+
+  configType = mkOptionType {
+    name = "nixpkgs-config";
+    description = "nixpkgs config";
+    check = x:
+      let traceXIfNot = c:
+            if c x then true
+            else lib.traceSeqN 1 x false;
+      in traceXIfNot isConfig;
+    merge = args: fold (def: mergeConfig def.value) {};
+  };
+
+  overlayType = mkOptionType {
+    name = "nixpkgs-overlay";
+    description = "nixpkgs overlay";
+    check = builtins.isFunction;
+    merge = lib.mergeOneOption;
+  };
+
+  _pkgs = import <nixpkgs> (
+    filterAttrs (n: v: v != null) config.nixpkgs
+  );
+
+in
+
+{
+
+  ###### interface
+
+  options = {
+
+    nixpkgs = {
+      config = mkOption {
+        default = null;
+        example = { allowBroken = true; };
+        type = types.nullOr configType;
+        description = ''
+          The configuration of the Nix Packages collection. (For
+          details, see the Nixpkgs documentation.) It allows you to set
+          package configuration options.
+
+          </para><para>
+
+          If <literal>null</literal>, then configuration is taken from
+          the fallback location, for example,
+          <filename>~/.config/nixpkgs/config.nix</filename>.
+
+          </para><para>
+
+          Note, this option will not apply outside your Nix-on-Droid
+          configuration like when installing manually through
+          <command>nix-env</command> or in your Home Manager config.
+
+          </para><para>
+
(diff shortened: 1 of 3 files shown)


modules/nixpkgs.nix
```diff
@@ -0,0 +1,137 @@
+# Licensed under GNU Lesser General Public License v3 or later, see COPYING.
+# Copyright (c) 2019 Alexander Sosedkin and other contributors, see AUTHORS.
+
+# Adapted from nixpkgs/home-manager.
```
t184256: Given it's MIT-licensed, you might have to embed the original license header as described in
http://softwarefreedom.org/resources/2007/gpl-non-gpl-collaboration.html,
section 2.2 "Adding GPL'd modifications to permissive-licensed files".

modules/nixpkgs.nix
```diff
@@ -0,0 +1,137 @@
...
+          </para><para>
+
+          If <literal>null</literal>, then configuration is taken from
+          the fallback location, for example,
+          <filename>~/.config/nixpkgs/config.nix</filename>.
+
+          </para><para>
+
+          Note, this option will not apply outside your Nix-on-Droid
+          configuration like when installing manually through
```
t184256: And it won't apply to home-manager configuration either. Should it? If not, should we probably state that too and provide a snippet to pass it through to it?

Same goes for overlays.

Gerschtli: Like in nixpkgs there is also no such comment, but I can add a simple note

modules/nixpkgs.nix
Gerschtli: Is there a possibility to add just a small one-/two-liner like your gpl header and add the complete license add the repo root? Because I would rather not paste 20 or more useless comment lines as this should be done for some more files already present in nix-on-droid.. But I never really got in touch with these license things.

Gerschtli:
I just added two new options to inject config and overlays in home-manager config

t184256:
Uhh, does this warrant separate options? I was hoping for some single-line trick in the doctext, this seems to be an overkill.

Gerschtli:
You mean one flag for inject config and overlays instead of two?

t184256:
No, I meant something like (warning: I'm totally making this up on the spot)

```
let
  overlays = ...
in
  {
   nixpkgs.overlays = overlays;
   home-manager.config.nixpkgs.overlays = overlays;
  }
```
```
let
  nixpkgs.overlays = ...
in
  {
    inherit nixpkgs.overlays;
    home-manager.config = {
      inherit nixpkgs.overlays;
    };
  }
```

or even just

```
rec {
  nixpkgs.overlays = ...
  nixpkgs.config = ...
  home-manager.config = {
    inherit nixpkgs;
 };
}
```

if anything of this makes any sense.

Gerschtli:
The Problem with this is, you can not use home-manager.config as attribute set, it is a home-manager module and therefore a function **or** an attribute set.. What we could do (or to be more clear: try) is to write custom option type for the hm config which injects those attributes into the config function, but I don't know how we could achieve this..

Or add an option like home-manager.configExtra which is an attribute set which we could inject into the home-manager config.

There are plenty of ways, the question is, what is the goal? Do we want a generic way to set arbitrary options in hm config or is nixpkgs.config/.overlays enough?

What are the up/downsides of your approach compared to mine?

t184256:
The Problem with this is, you can not use home-manager.config as attribute set, it is a home-manager module and therefore a function or an attribute set..

Sorry, I don't get it. I'm not trying to use it as an attribute set, I'm trying to inject values into it with `inherit`.

```
nix-repl> ((let nested.attr = "value"; in rec { func = param: {inherit nested;}; } ).func "arg").nested.attr
"value"
```

> There are plenty of ways, the question is, what is the goal?

I wanted to just add a hint to the doc: 'to set these values inside h-m config, you can access the outer ones in this manner'.

> What are the up/downsides of your approach compared to mine?

Pros:

* flexible
* clear
* no options required, you just do what you meant to do: set this to that using existing Nix constructs

Cons:

* no idea if this is possible
* I clearly cannot communicate my idea

Will try it later and post a working config snippet if I manage to produce it.

t184256:
Passing the fixed point of the outer config works:
```
{pkgs, config, ...}:
{
  nixpkgs.overlays = [ (self: super: { unstable = import <nixos-unstable> {}; })];
  home-manager.config = {pkgs, ...}:
  {
    inherit (config) nixpkgs;
  };
}
```
The only downside I see is potential shadowing by the `config` fixed point of the h-m config. Will think more.

t184256:
```
{pkgs, config, ...}:
{
  nixpkgs.overlays = [ (self: super: { unstable = import <nixos-unstable> {}; })];
  home-manager.config = let outer-config = config; in {pkgs, config ...}:
  {
    # now both outer-config and config are available, one can do, e.g.,
    inherit (outer-config) nixpkgs;
  };
}
```
Anything wrong with that one?

Gerschtli:
Yes I agree, this works, but only if you define the `home-manager.config` inline. In my setup (https://github.com/Gerschtli/dotfiles/blob/master/nix-on-droid/nix-on-droid.nix) I import my `home.nix` file. There I don't have access to the outer config. We could add something like `nix-on-droid-config` as parameter for each `home-manager` module via `_module.args`.

Gerschtli:
Somehow this works:

```nix
{ pkgs, lib, config, ...}:

lib.mkMerge [
  {
    home-manager.config.nixpkgs.config = import ./nixpkgs-config.nix;
    nixpkgs.config = import ./nixpkgs-config.nix;
  }
  {
    home-manager.config =
      { pkgs, ... }:
      {
        # other config
      };
  }
]
```

So I removed the inject options and updated the option descriptions.

t184256:
Isn't this just

```
{ pkgs, lib, config, ...}:

  {
    home-manager.config.nixpkgs.config = import ./nixpkgs-config.nix;
    nixpkgs.config = import ./nixpkgs-config.nix;
    home-manager.config =
      { pkgs, ... }:
      {
        # other config
      };
  }
```

?

Gerschtli:
You cannot set an option twice AFAIK. Therefore the mkMerge in my example.

t184256:
Automatically merging values from multiple assignments into attribute sets is perfectly, but now I see how you're merging an attrset with a function-returning-an-attrset, and I have no idea what's going on here.

Gerschtli:
This is the part where I have no clue what is happening, but it works.. :D I'm open for better ideas but I think for now we can leave it like this as we are not defining any new options.

t184256:
Yeah, you are right. My objection was adding extraneous options, and now it's fine to land this without figuring out the perfect docstring from the first attempt.

PR #42, opened 2020-01-07
home-manager: Add backupFileExtension option

Gerschtli:
Adds new option to be able to set `HOME_MANAGER_BACKUP_EXT` (see `-b` flag for home-manager) via nix-on-droid config.


t184256:
Worked fine, I'm merging it.

PR #43, opened 2020-01-07
time: Add timeZone option

Gerschtli:
Add option for time zone.

Resolves https://github.com/t184256/nix-on-droid-bootstrap/issues/25


t184256:
force-pushed to solve a trivial conflict in the module list

PR #45, opened 2020-01-08
Host name

Gerschtli:
I tried to add an option for setting the host name, but somehow it does not work like expected. Do you have an idea?


t184256:
My understanding is that hostname is `/proc/sys/kernel/hostname`, and `/etc/hostname` is just a way to persist it across reboots.

Gerschtli:
Ah this seems a lot tricker than I thought.. I would like to postpone this and open a separate issue for it when I got more time for it

PR #46, opened 2020-01-08
login-inner: source session-init on initial build

Gerschtli:
Source session-init on initial build which closure is added in `nix-directory`.

Can you test it? I don't want to wipe my data again.. :D


t184256:
Sure. Now installation passes, /proc warning are gone, `env` seems alright before and after re-login. Will merge.

PR #47, opened 2020-01-08
nix-on-droid: add -I argument

Gerschtli:
Pass -I argument to nix-build.

Closes https://github.com/t184256/nix-on-droid-bootstrap/issues/40

### DIFF
--- a/nix-on-droid/nix-on-droid.sh
+++ b/nix-on-droid/nix-on-droid.sh
@@ -25,10 +25,11 @@ function doHelp() {
     echo "  -h|--help       Print this help"
     echo "  -n|--dry-run    Do a dry run, only prints what actions would be taken"
     echo "  -v|--verbose    Verbose output"
-    echo "  -f|--file       Path to config file"
+    echo "  -f|--file FILE  Path to config file"
     echo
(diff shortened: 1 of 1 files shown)


Gerschtli:
Ah yes, the `<nix-on-droid>` one needs the `NIX_PATH` info, will fix it tomorrow

Gerschtli:
Ah wait a second: my changes are working as expected, you missed the `nix-on-droid=` prefix in your command:

```diff
-nix-on-droid -I $(pwd) switch
+nix-on-droid -I nix-on-droid=$(pwd) switch
```

t184256:
Yeah, you're right, my bad. Seems to work OK if used properly.

And now it feels weird that the repository name isn't 'nix-on-droid'. Hope it won't be too confusing later.

Gerschtli:
Maybe we should consider renaming it because the name does not longer match the content..

As this is still in a very early stage, better do it now than later. Don't know how to let the user know that the channel url has changed. Maybe redirect old channel url to the new one.

t184256:
Seems like it's a pretty straightforward low-risk thing, huh. Feeling courageous today, renamed.

PR #49, opened 2020-01-13
Add deploy helper script

Gerschtli:
I added a little helper script to build and deploy the bootstrap zip ball and optional a custom channel tar ball.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,5 +8,5 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
-[*.sh]
+[{scripts/deploy,*.sh}]
 indent_size = 4
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 result
 result-*
-out
+scripts/config.sh
--- a/README.md
+++ b/README.md
@@ -152,6 +152,25 @@ In case you only care about updates through wiping the data,
 you shouldn't need a binary cache for that.
 
 
+### `scripts/deploy` helper script
+
+To simplify the local development and deployment there is a `scripts/deploy`:
+
+```
+$ ./scripts/deploy help
+Usage: ./scripts/deploy [COMMAND]
+
+Commands (default: all)
+
+  all         Build and upload bootstrap zip ball and channel if enabled
+  bootstrap   Build and upload bootstrap zip ball
+  channel     Build and upload channel tar ball if enabled
+  help        Print this help
+```
+
+Some static config needs to be set in `scripts/config.sh` (see
+`scripts/config.sh.dist`).
+
 
 ## Tips
 
--- a/scripts/config.sh.dist
+++ b/scripts/config.sh.dist
@@ -1,0 +1,13 @@
+# rsync destination path
+#COPY_TO="host.name:/path/to/directory/"  # with trailing slash
+
+# url to custom nix-on-droid channel, either via github like:
+#NIX_ON_DROID_CHANNEL="https://github.com/<USER>/<REPO>/archive/<BRANCH>.tar.gz"
+# or via upload to $COPY_TO like:
+#NIX_ON_DROID_CHANNEL="https://example.com/path/to/"  # with trailing slash, "channel.tar.bz2" will be appended
+#UPLOAD_CUSTOM_CHANNEL=1
+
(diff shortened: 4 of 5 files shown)


t184256:
Why bother building channel tarballs?

Gerschtli:
Because it is way faster and easier than commit to a temporary branch and push to github. And I like to develop on my laptop instead of my phone.

t184256:
IDK, I'm not a fan. Here's [what I use](https://gist.github.com/t184256/33e80eb1a48542b58ce5443864ccf8ca). It's half the size, has a comparable feature set and doesn't require a config file. I'd say that the url to push to / .bz2 channel url are perfectly fine as an envvar or a command-line argument, while the other stuff is autodetectable anyway.

The other thing that turns me off is packing the entire current working directory as a channel tarball. Mine's not that tidy.

Finally, my script was mostly suggested as a way to guide a potential forker (as I already had a command in my shell history that does effectively the same). Your version seems more confusing to me, but I'm biased, of course.

Gerschtli:
My goal with this script was to unify the scripts that we both already use and provide a reasonable way to configure possible parameter. Because why should a potential contributor rebuild such a script if we already have working ones?

The way it is configured can be changed of course, but my idea behind a config file was that the frequently use of `./scripts/deploy` is as simple as possible regarding possible forgotten command line arguments, but I am open for ideas.

> while the other stuff is autodetectable anyway

What do you mean with "other stuff"? I tried to use as few config parameters as possible.

> The other thing that turns me off is packing the entire current working directory as a channel tarball. Mine's not that tidy.

Can you elaborate what you mean by that? To be exact, there are three possible workflows to develop a new module:

1. Clone the repo on the device and use the `-I nix-on-droid=/path/to/repo` argument
2. Develop on another device, commit changes, push to github and update the nix-channel
3. Develop on another device, create tar ball, upload it and update the nix-channel

Maybe there are more ways, but I think these three are perfectly reasonable ways to achieve the same goal.

BTW: The tar balls used in 2 and 3 are exactly the same.

> Your version seems more confusing to me

What is confusing you? Is the feature set of this script not clear or do you need more documentation?

Or to sum up all of this: Is it, that you generally dislike the idea of a script like that?

Gerschtli:
FYI: This is the script I am using to build and push the bootstrap zip ball and the channel tar ball:

```bash
#!/usr/bin/env bash
set -euo pipefail

cd "$(dirname "$0")"

nix build -f pkgs --argstr arch aarch64 --argstr nixOnDroidChannelURL "https://downloads.tobias-happ.de/bootstrap/channel.tar.bz2" bootstrapZip
tar --exclude=".git" --exclude="channel.tar.bz2" --exclude="result" --transform 's,^,nix-on-droid/,' -cjf channel.tar.bz2 .
scp result/bootstrap-aarch64.zip channel.tar.bz2 host.name:/path/to/bootstrap
```

Maybe the `scripts/deploy` can be simplified.

t184256:
OK, I see that my comments were pretty confusing.

By "packing the entire current working directory as a channel tarball" I meant that it also packed files not managed by git.

I like the idea of having such a script, but I think that it should be less of a polished CLI tool and more of a literate programming-style guide on what on Earth is going on here and what are we automating in the first place.
I've started rewriting my script, but I'm short of time, as always =/

Gerschtli:
> By "packing the entire current working directory as a channel tarball" I meant that it also packed files not managed by git.

But that is the point of it :D I want to develop on new modules -> new files -> not tracked by git. I don't see anything bad with it.

> I like the idea of having such a script, but I think that it should be less of a polished CLI tool and more of a literate programming-style guide on what on Earth is going on here and what are we automating in the first place.

So you would prefer to list and explain the commands in the Readme instead of a script?

t184256:
> But that is the point of it :D I want to develop on new modules -> new files -> not tracked by git.

ooh. right. silly me.

> So you would prefer to list and explain the commands in the Readme instead of a script?

[Here's what I started writing last night](https://gist.github.com/t184256/c79d270498e8441d0f4e3c9be7acad18). It stops making sense past some point, but you get the idea about the format. Something between a maintainer doc with bits of working code.

Gerschtli:
As I'm currently developing through ssh, my PR seems to be an overkill but feel free to reopen.

PR #50, opened 2020-01-14
Add overlays for htop and proot-termux

Gerschtli:
Hey, I am working on some overlays for htop for example. This works already, so I tried building proot on the device but if I run `nix-channel --update` I got an error like `Bus error` with an exit code of 135.

Do you have any idea, what might cause this?

Furthermore to provide a nice user experience, we need to build and push these packages with overlays to cachix. This is the reason for the pinned nixpkgs in the overlays to manually update the nixpkgs version to prevent rebuilds.

If an aarch64 build machine like in travis or github workflow really builds incompatible packages, we should consider using a custom android phone as build machine via hercules-ci or github workflow self-hosted runner.

What do you think?

### DIFF
--- a/modules/environment/login/default.nix
+++ b/modules/environment/login/default.nix
@@ -82,14 +82,7 @@ in
     environment.files = {
       inherit login loginInner;
 
-      prootStatic =
-        let
-          crossCompiledPaths = {
-           aarch64 = "/nix/store/r09n7pp4fwhrld2a1k2al6bgdx2qqfaj-proot-termux-unstable-2019-09-05-aarch64-unknown-linux-android";
-           i686 = "/nix/store/4gpm0rmrq0mm69kl3cb1gjslr7ihhp01-proot-termux-unstable-2019-09-05-i686-unknown-linux-android";
-          };
-        in
-          "${crossCompiledPaths.${config.build.arch}}";
+      prootStatic = pkgs.proot-termux;
     };
 
   };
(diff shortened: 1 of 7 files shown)


t184256:
Lumping these two into a single PR wasn't probably a good idea, as the complexity of these two tasks is super-disparat.

Any journey towards building a functioning proot that don't rely on cross-compiling it with NDK need to start with a proof of concept. I want to first know it's possible, and only then polish the way of pulling it off with Nix and figure out the required infrastructure. #1 and #6 are related. If I were somehow super-motivated to build it some other way, I'd probably roll back a-a-all the way back to https://github.com/t184256/nix-in-termux for experimenting and started with building vanilla proot on-device inside Termux until I have something free of those mysterious bus errors. This is definitely something wilder and more complicated than shipping an overlay.

Heck, even going custom-prefix and ditching proot could be a simpler task in the end. At least it's clear what to do.

t184256:
I've tested `htop`, found it working (great!) and fast-forwarded master to 93d081c. So, half-merged.

t184256:
Ugh. Accidentally un-marked it as draft and now I can't find how to undo that =/ Sorry.

Gerschtli:
> Lumping these two into a single PR wasn't probably a good idea, as the complexity of these two tasks is super-disparate.

You are right, I should have split this. Thank you for merging the htop patch anyway :)

To explain myself on the proot overlay: I thought you had already be able to compile proot natively on an android device so I thought I had make a mistake, hence the PR. As this assumption was wrong, I will close this PR.

PR #57, opened 2020-04-23
Add support for different login shells

Gerschtli:
This change allows to change the login shell through providing the full path to the binary.

In case the binary is not present or executable, bash will be run instead.

Through the failsafe shell it is possible to override the passwd and change the shell manually

### DIFF
--- a/modules/environment/login/login-inner.nix
+++ b/modules/environment/login/login-inner.nix
@@ -69,6 +69,7 @@ writeText "login-inner" ''
   elif [ -x "${config.user.shell}" ]; then
     exec "${config.user.shell}"
   else
+    echo "Cannot execute shell '${config.user.shell}', falling back to bash"
     exec /usr/bin/env bash
   fi
 ''
--- a/modules/environment/login/login-inner.nix
+++ b/modules/environment/login/login-inner.nix
@@ -11,9 +11,10 @@ writeText "login-inner" ''
 
   set -eu -o pipefail
 
-  [ "$#" -gt 0 ] || echo "Welcome to Nix-on-Droid!"
-
-  [ "$#" -gt 0 ] || echo "If nothing works, open an issue at https://github.com/t184256/nix-on-droid/issues or try the rescue shell."
+  if [ "$#" -eq 0 ]; then  # if script is called from within nix-on-droid app
+    echo "Welcome to Nix-on-Droid!"
+    echo "If nothing works, open an issue at https://github.com/t184256/nix-on-droid/issues or try the rescue shell."
+  fi
 
   ${lib.optionalString config.build.initialBuild ''
     if [ -e /etc/UNINTIALISED ]; then
@@ -63,9 +64,11 @@ writeText "login-inner" ''
   . "${config.user.home}/.nix-profile/etc/profile.d/nix-on-droid-session-init.sh"
   set -u
 
-  if [ "$#" -eq 0 ]; then
-    exec /usr/bin/env bash
-  else
+  if [ "$#" -gt 0 ]; then  # if script is not called from within nix-on-droid app
     exec /usr/bin/env "$@"
+  elif [ -x "${config.user.shell}" ]; then
+    exec "${config.user.shell}"
+  else
+    exec /usr/bin/env bash
   fi
 ''
--- a/modules/user.nix
+++ b/modules/user.nix
@@ -40,7 +40,7 @@ in
 
       shell = mkOption {
         type = types.path;
-        readOnly = true;
+        default = "${pkgs.bashInteractive}/bin/bash";
         description = "Path to login shell.";
       };
 
@@ -73,7 +73,6 @@ in
     user = {
       group = "nix-on-droid";
       home = "/data/data/com.termux.nix/files/home";
-      shell = "/bin/sh";
       userName = "nix-on-droid";
     };
 
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -5,12 +5,16 @@
 let
   loadNixpkgs = import lib/load-nixpkgs.nix;
 
+  nixDirectory = callPackage ./nix-directory.nix { };
+  packageInfo = import "${nixDirectory}/nix-support/package-info.nix";
+
   nixpkgs = loadNixpkgs { };
 
   modules = import ../modules {
     pkgs = nixpkgs;
 
     config = {
+      user.shell = "${packageInfo.bash}/bin/bash";
       imports = [ ../modules/build/initial-build.nix ];
 
       _module.args = { inherit customPkgs; };
@@ -34,10 +38,9 @@ let
   );
 
   customPkgs = rec {
+    inherit nixDirectory packageInfo;
     bootstrap = callPackage ./bootstrap.nix { };
     bootstrapZip = callPackage ./bootstrap-zip.nix { };
-    nixDirectory = callPackage ./nix-directory.nix { };
-    packageInfo = import "${nixDirectory}/nix-support/package-info.nix";
     prootTermux = callPackage ./cross-compiling/proot-termux.nix { };
     qemuAarch64Static = callPackage ./qemu-aarch64-static.nix { };
     tallocStatic = callPackage ./cross-compiling/talloc-static.nix { };


modules/environment/login/login-inner.nix
```diff
@@ -76,9 +76,10 @@ writeText "login-inner" ''
   # Workaround for https://github.com/NixOS/nix/issues/1865
   export NIX_PATH=nixpkgs=$HOME/.nix-defexpr/channels/nixpkgs/:$NIX_PATH
 
-  if [ "$#" -eq 0 ]; then
-    exec /usr/bin/env bash
+  __NOD_SHELL="$(getent passwd ${config.user.userName} | cut -d : -f 7)"
```
t184256, line 79: Leads to `getent: command not found` for me for

1) post-installation (logging in again helps)
2) post-shell-change (fatal)

Gerschtli: Oh thats strange.. It did work for me, let me investigate it a bit.

Gerschtli: Removed the getent call completely. Was unneccessary on second look ;)

Gerschtli: But can you tell me, what this if condition means?

`if [ "$#" -eq 0 ]; then`

t184256: 'if called with zero arguments'

Gerschtli: Yes, that is clear to me, but what does is mean in this context? Under which circumstances are the arguments set? Why did you choose to use this condition in the first place?

t184256: Ahh, right. That's a very important (to me) undocumented feature.

`login` (and `login-inner`) can be called directly from other parts of Android to run something that's managed by nix-on-droid. Then login enters proot, the command gets plumbed up to this place, the path to the command would resolve to the one valid inside proot, and, finally, it'll get get executed with the provided arguments.

Please keep it, I need it for my Bluetooth keyboard remapper. =)

modules/user.nix
```diff
@@ -40,7 +40,7 @@ in
 
       shell = mkOption {
         type = types.path;
-        readOnly = true;
+        default = "/bin/sh";
```
t184256, line 43: I think I've written this before, but I don't see this comment, so sorry if it'll be a duplicate.

Why `/bin/sh`? Isn't it a user experience regression? Can it somehow default to `${pkgs.bash}/bin/bash`?

modules/environment/login/login-inner.nix
Gerschtli: Ah okay, I will add a small comment in there to explain it

modules/user.nix
Gerschtli: Yes of course, you are right, I'll update this

Gerschtli:
Changed it to bashInteractive as default shell and added the if condition back in

t184256:
I've force-pushed two commits, one highlighting that shell is set incorrectly on the initial build and one workarounding it. Are you fine with this workaround?

Gerschtli:
You may update the comment for the initialBuild workaround as I can explain you exactly why this isn't working: On initial build, the path to bashInteractive is not present in the nix store.

A better way to fix this, may be to update `pkgs/nix-directory.nix` and add `config.user.shell` in `rootPaths` of `prootTermuxClosure`. Could you test this? Because I currently can't reset my device.

```diff
--- a/pkgs/nix-directory.nix
+++ b/pkgs/nix-directory.nix
@@ -18,6 +18,7 @@ let
 
   prootTermuxClosure = closureInfo {
     rootPaths = [
+      config.user.shell
       prootTermux
     ];
   };
```

t184256:
> You may update the comment for the initialBuild workaround as I can explain you exactly why this isn't working: On initial build, the path to bashInteractive is not present in the nix store.

I admit I don't get why. Initial build ends with a switch, switch should've brought it to nix store. What am I missing?

> A better way to fix this... Could you test this?

Sure, will do.

Gerschtli:
Yes it does a switch, with a completely other nixpkgs revision which results in a different store path.

t184256:
Ah, right. And we're already executing the old login-inner.

I'm hitting `proot info: vpid1: terminated with signal 4` for some reason (hope it's unrelated), will debug further.

t184256:
I'm reasonably sure that change puts a host-arch bash into the zipball, crashing the installation at the very last step.

Gerschtli:
Oh, I missed that.. The last idea to fix this properly would be setting the `config.user.shell` in `pkgs/default.nix` line 13 to `${packageInfo.bash}/bin/bash` if it does not produce an infinite recursion.. If this does not work, I'm fine with your workaround.

t184256:
That seemed to work!

Gerschtli:
Nice! That I'm fine with this PR if you'd like to merge :)

t184256:
Merging, thanks!

Shame my favourite shell (xonsh) is embarrassingly slow on my tablet, otherwise I'd use that feature.

Narice:
Hello! Thank you for this feature!
There is one thing that does not work for me however which is the fallback to bash if the other shell is not working
Basically, I changed the default value to `pkgs.zsh` which is a very incorrect syntax for what I wanted to do, and when starting a new session, the session exits with this message:
```
Welcome to Nix-on-Droid!
If nothing works, open an issue at https://github.com/t184256/nix-on-droid/issues or try the rescue shell.
/data/data/com.termux.nix/files/usr/usr/lib/login-inner: line 21: /nix/store/cd9rscw4w3cvzwq092i2cikfn168kv87-zsh-5.7.1: Is a directory
/data/data/com.termux.nix/files/usr/usr/lib/login-inner: line 21: exec: /nix/store/cd9rscw4w3cvzwq092i2cikfn168kv87-zsh-5.7.1: cannot execute: Is a directory

[Process completed (code 126) - press Enter]
```
From the rescue shell, it is simppy not possible to recover from this as I can't use `nix-on-droid switch`
Even when going into the store and trying to launch it from there (doing a `./nix-on-droid switch` results in a `sh: ./nix-on-droid: No such file or directory`
I hope this helps the project!

Gerschtli:
Hey, so you should have used
```
user.shell = "${pkgs.zsh}/bin/zsh";
```
in your config in the first place.

To roll back you can use the rescue shell and run the activate script in a previous build of `/nix/var/nix/profiles/nix-on-droid`. In theory this should be everything you need to restore a previous generation.

t184256:
... prefixed with `login` script, right?

Actually, can't you just `/data/data/com.termux.nix/files/usr/bin/login bash`?

Gerschtli:
What do you want to prefix with login script?
And where do you want to run that command?

t184256:
In the rescue shell; to get into the proot environment.

Gerschtli:
Ah yes of course you are right. That should also work

Narice:
Oh woaw thanks guys!
I thought my nix-on-droid was deadlocked
Thank you for the tips!

mhuesch:
apologies if I should use a different place to ask this, but I have the same issue that Narice had. yet, I am unsure what "rescue shell" means and I have searched this repo and not turned up anything...

how does one launch the rescue shell if `termux.nix` app is unable to start (because of bad login shell path)?

t184256:
@mhuesch Stop all sessions (e.g., using the notification), ensure Nix-on-Droid android app isn't running. In the application drawer, pick the variant of the app with the pale icon, named "Nix (failsafe)" or something like this. You'll get a regular Android environment shell from the same user Nix-on-Droid uses, allowing you to perform recovery activities.

mhuesch:
@t184256 found it, thanks so much!

really appreciate the work you (& others) have done on this project, it has become an integral part of my android/linux shared workflow :bow: :pray:

t184256:
Thanks for the kind words.

mags-issues:
So, ehm, I wanted to avoid bumping this, but I've had bad luck with my computing lately and I can't with rescue shells today.

Where exactly do I put this: `user.shell = "${pkgs.zsh}/bin/zsh";`?

In `~/.config/nix-on-droid/nix-on-droid.nix`? Feels like it should go in the home-manager setup, but idk if that's possible. My brain is boiling with new concepts 🫠  Thank you for a wonderful project though! Amazing stuff 🙏 💖

t184256:
Into nix-on-droid, it's a [nix-on-droid option](https://nix-community.github.io/nix-on-droid/nix-on-droid-options.html#opt-user.shell). Just like home-manager doesn't manage logging into the user on NixOS or Linux+Nix, it doesn't manage it on nix-on-droid. In general, it might be easier to start with much more popular Linux+Nix or NixOS and only then move to nix-on-droid, otherwise you're playing the Nix game on "hard" difficulty.

mags-issues:
Thank you 🙏 I wish I could, but I'm on the road and all I've got is Android + hw keyboard 💁‍♀️

PR #58, opened 2020-04-23
Add source nix.sh and hm-session-vars.sh in session-init.sh

Gerschtli:
With this change, everything to setup the session is separated into the session-init.sh. This has the advantage that when logged in via ssh, you only need to source one file, instead of up to 3.

### DIFF
--- a/modules/environment/login/login-inner.nix
+++ b/modules/environment/login/login-inner.nix
@@ -4,11 +4,6 @@
 
 let
   inherit (customPkgs.packageInfo) cacert coreutils nix;
-
-  sessionInitPackage =
-    if config.build.initialBuild
-    then config.build.sessionInit
-    else "/nix/var/nix/profiles/per-user/${config.user.userName}/profile";
 in
 
 writeText "login-inner" ''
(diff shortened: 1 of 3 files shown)


modules/environment/login/login-inner.nix
```diff
@@ -62,19 +59,9 @@ writeText "login-inner" ''
...
-    [ "$#" -gt 0 ] || echo "Sourcing home-manager environment..."
-    export NIX_PATH=$HOME/.nix-defexpr/channels''${NIX_PATH:+:}$NIX_PATH
-    set +u
-    . "$HOME/.nix-profile/etc/profile.d/hm-session-vars.sh"
-    set -u
-  fi
-
-  # Workaround for https://github.com/NixOS/nix/issues/1865
-  export NIX_PATH=nixpkgs=$HOME/.nix-defexpr/channels/nixpkgs/:$NIX_PATH
+  set +u
```
t184256, line 62: I think this should be moved inside the file.

Gerschtli: I think it does not belong in the file because the `set +u` and `set -u` only make sense if the script that sources the file has `set -u`, otherwise that could result in unexpected behaviour.

t184256: True, and saving/restoring the options, while possible, isn't nice. Fine, no more comments then.

t184256:
Will do upgrade testing later and merge.

PR #59, opened 2020-04-23
home-manager 20.03 upgrade

Gerschtli:
This PR introduces changes needed for the 20.03 release of home-manager. On 19.09 releases of home-manager this won't work.

Furthermore I added the useGlobalPkgs option recently introduced to home-manager.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ for example:
 
 {
   environment.packages = [ pkgs.vim ];
-  system.stateVersion = "19.09";
+  system.stateVersion = "20.03";
 }
 ```
 
@@ -66,7 +66,7 @@ An alternative location is `~/.config/nixpkgs/config.nix` with the key
 
     {
       environment.packages = [ pkgs.vim ];
-      system.stateVersion = "19.09";
+      system.stateVersion = "20.03";
     };
 }
 ```
@@ -80,7 +80,7 @@ To enable `home-manager` you simply need to follow the instructions already prov
 
 1.  Add `home-manager` channel:
     ```
-    nix-channel --add https://github.com/rycee/home-manager/archive/release-19.09.tar.gz home-manager
+    nix-channel --add https://github.com/rycee/home-manager/archive/release-20.03.tar.gz home-manager
     nix-channel --update
     ```
 2.  Configure `home-manager`:
--- a/default.nix
+++ b/default.nix
@@ -38,8 +38,8 @@ rec {
       echo
       echo "Recommended: update with what you've used before (home-manager switch -or- nix-env -u)"
       echo
-      echo "Required if you've followed the steps above: set nix-on-droid channel to master"
-      echo "  nix-channel --add https://github.com/t184256/nix-on-droid/archive/master.tar.gz nix-on-droid"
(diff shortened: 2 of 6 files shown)


t184256:
Does that mean that it'll stop working with the 19.09 one?

Gerschtli:
Yes, thats right unfortunately.. They updated the requirements for the modules.nix args.

t184256:
Then the minimum is probably to update the home-manager version we setup for the new installs, the reasonable maximum is to block updates with a meaningful message when the home-manager version is too old.

Gerschtli:
Or we could go the way like home-manager and create separate channels for nixos releases.

t184256:
That's a very good idea, and I'm all for it, but it's orthogonal to the problem of upgrading the existing installations.

Gerschtli:
This PR does now contain the https://github.com/t184256/nix-on-droid/pull/60.

So if you create a `release-19.09` and `release-20.03` branch we could advise the users to use the release channels in the future.

t184256:
I've created/pushed a `release-19.09` branch, and I'm going to build/publish the corresponding bootstraps as well. Will create a `release-20.03` on merging this one.

Could you please add something like an assertion for the presence of that new option, with a message instructing to use a newer h-m channel?

Gerschtli:
Something like this? You should consider removing this hack in the near future as it does not allow config merges for the home-manager options..

t184256:
Tried upgrading to 8cc3540 at my testing device, got
`file 'home-manager/modules/modules.nix' was not found in the Nix search path ... modules/home-manager.nix:10:45`.

BTW, any other upgrading ideas aside of adding that assertion?

Gerschtli:
Could you try this one?

t184256:
The assertion fired up OK. Switched the channel, rebuild fails with file `attribute 'submoduleWith' missing, at .../modules/home-manager.nix:15:14`.

t184256:
Changing the nixpkgs to 20.03 made it go away. Is it worth reflecting in the same/separate assertion message?

Gerschtli:
Oh yes, for sure. Just make sure the commits with the assertions are not added in release-20.03.

I can write the assertion tomorrow or you could do it if you like.

t184256:
> Just make sure the commits with the assertions are not added in release-20.03.

Why? And how? They have to be on master.

t184256:
OK, I'd like to leave it to you, I guess.

Just FYI, I went ahead and

1) tested that the clean installation works: it does
2) updated my 'stable' tablet to 20.03: it worked well too

So, all in all, it's working fine, great job! I'd merge it already if not for the second assertion and that comment of yours about not having them on 20.03. It can wait, no rush.

Gerschtli:
I've added the assertion.

For my other comment: I thought about it, but as everyone currently uses master as nix-channel, its alright. When 20.09 is out we should maybe drop the assertions and fully use the release branches for this kind of assertions

t184256:
Re-tested upgrading once again, worked as intended. Thanks a lot for your work!

Gerschtli:
Fun as always :)

PR #60, opened 2020-04-23
Add 20.03 stateVersion

Gerschtli:


t184256:
What's the point till these are equal?

Gerschtli:
If we provide an stateVersion option, we should allow setting it to the most recent. stateVersion is just to keep track of the state and if we dont need it yet, it is totally fine.

This would be analogous to the behaviour of nixos and home-manager

t184256:
OK, why not. Closing as it's going to come as part of #59.

PR #73, opened 2020-08-08
Add `useUserPackages` documentation

ShamrockLee:
Add `#home-manager.useUserPackages`
and explanation
to `nix-on-droid.nix.default` and `README.md`

If applied, users will be able to resolve package conflicts and make `home-manager` work more easily.

Solve issue #65

PR #90, opened 2020-11-01
Add release-20.09

Gerschtli:
I prepared everything for a branch off for release-20.09.

Things to consider:
- [ ] Adding a changelog?
- [ ] Updating nixpkgs version for bootstrap and overlays?
- [ ] Maybe use `qemu` and `talloc` packages from nixpkgs?
- [ ] Define an EOL for the migration logic in `default.nix` and previous releases?

Fixes https://github.com/t184256/nix-on-droid/pull/73.


t184256:
Two gifts in one: an execution and a plan! I've followed along a bit:

* Adding a changelog: a436884
* Updating nixpkgs version for bootstrap and overlays: 06db83b
* Maybe use qemu and talloc packages from nixpkgs: for talloc see a07c249, for qemu... not worth the hassle, it's used for a fraction of a second, while compilation will take ages
* Define an EOL for the migration logic in default.nix and previous releases: drop it =)

default.nix
```diff
@@ -73,44 +73,48 @@ rec {
...
               -  # After installing home-manager channel like
-              -  #   nix-channel --add https://github.com/rycee/home-manager/archive/release-20.03.tar.gz home-manager
+              -  #   nix-channel --add https://github.com/rycee/home-manager/archive/release-20.09.tar.gz home-manager
               -  #   nix-channel --update
               -  # you can configure home-manager in here like
               -  #home-manager.config =
               -  #  { pkgs, ... }:
               -  #  {
+              -  #    # Read the changelog before changing this value
+              -  #    system.stateVersion = "20.09";
```
t184256, line 88: What is this one? Did you mean home.stateVersion?

Gerschtli: Oh, you are right, I meant home.stateVersion :) You can change it if you like, otherwise I will do it in the next days

Gerschtli:
Thank you, your changes look good to me, maybe add a link to the mentioned upgrade instructions in the Changelog. And if we want to drop support for the migration, we should remove it from default.nix too?

t184256:
Another question that comes to me each release: what's the acceptable threshold of successfully buildable aarch64 packages?

Gerschtli:
First, we should fix my mistake concerning home.stateVersion in modules/environment/login/nix-on-droid.nix.default.

And for your question: I don't think it would be reasonable to be blocked by build failures of nixpkgs because nix-on-droid does not depend on all of these packages or to put it in other words: to fix these issues one would create a PR towards nixpkgs which will be released independently.

We should probably test that all mandatory packages for the nix-on-droid env build successfully but I think this will be always the case.
And we should in fact test that all packages influenced by our overlays build successfully (and push these to cachix).

t184256:
> First, we should fix my mistake concerning home.stateVersion in modules/environment/login/nix-on-droid.nix.default.

Amended.

> And for your question: I don't think it would be reasonable to be blocked by build failures of nixpkgs
> because nix-on-droid does not depend on all of these packages or to put it in other words:
> to fix these issues one would create a PR towards nixpkgs which will be released independently.
> 
> We should probably test that all mandatory packages for the nix-on-droid env build successfully
> but I think this will be always the case. And we should in fact test that all packages influenced
> by our overlays build successfully (and push these to cachix).

Yeah, that'll happen for sure.

I mean, from the separation of concerns, you're right.
But for a passerby from F-Droid who just wants ghc on their phone, releases are broken for a while.

All in all, I don't see any problems with the changes. Will redo clean install, migrate my setup and report back.

t184256:
Brief update: things aren't going too smoothly, gnumake 4.3 might be broken. Need more time investigating.

t184256:
There's a weird thing I observe on my updated-since-forever Android 9 tablet:

`make` fails to spawn any processes during on-device builds if it's gnumake 4.3. gnumake 4.2 works fine, but rebuilding nixpkgs doesn't sound any good to me. what's worse and weirdest, reverting to a pre-update generation doesn't fix the problem.

And I can't replicate the issue on an Android 8 phone, updated or clean-installed. So I hope it's just a false alarm and something that'd go away on a clean install, but I'd appreciate extra data points.

Another thing that has to be done is updating the overlays, but I'm blocked on a broken make ATM.

t184256:
It's still there even after the reinstall, and I think it could be the `posix_spawn` change. I'll file a separate issue if I don't figure it out in reasonable time.

Gerschtli:
Alright, thank you for your investigation. I'm not that into the C ecosystem but I will test it on my phone at the weekend.

t184256:
Opened a ['what do I even do' thread](https://discourse.nixos.org/t/nix-on-droid-gnumake-4-3-broke-how-do-i-graft-stdenv-what-else-can-i-do/9875) on discourse.nixos.org.

Gerschtli:
Building gnumake and gnumake42 both worked on my phone.

nix-info:
• system: "aarch64-linux" 
• host os: Linux 4.4.205-perf+ 
• multi-user?: no 
• sandbox: no 
• version: nix-env (Nix) 2.3.7 
• channels(nix-on-droid): "home-manager-20.09, nix-on-droid, nixpkgs-20.09.1599.d105075a1fd, nur-gerschtli, unstable-21.03pre250162.34ad166a830" 
• nixpkgs: /data/data/com.termux.nix/files/home/.nix-defexpr/channels/nixpkgs

t184256:
Ready for review!

modules/workaround-make.nix
```diff
@@ -0,0 +1,68 @@
...
+  config = {
+    build.extraProotOptions =
+      lib.optional (config.system.workaround.make-posix-spawn.enable)
+        (
+          "-b " +
+          "${config.build.installationDir}/" +
+            "${gnumake-no-posix-spawn}/bin/overlaying-make" +
+          ":" +
+          "${pkgs.gnumake}/bin/make"
+        );
```
Gerschtli, line 66: To be a bit more concise.

```suggestion
      optionals cfg.enable [
          "-b"
          (
            config.build.installationDir
            + "${gnumake-no-posix-spawn}/bin/overlaying-make"
            + ":${pkgs.gnumake}/bin/make"
          )
        ];
```
And in the top of this module define
```nix
let cfg = config.system.workaround.make-posix-spawn;
```

modules/workaround-make.nix
```diff
@@ -0,0 +1,68 @@
...
+        get stuck with a broken build; stop attempting to build something;
+        enable the option; nix-on-droid switch, relogin, build,
+        disable the option; nix-on-droid switch, relogin.
+
+        If you leave it on Nix-store validation will fail,
+        repairs will break the working make,
+        updates will do bad things. You have been warned.
+
+        Consider building remotely as an alternative for such devices:
+        https://github.com/t184256/nix-on-droid/wiki/Remote-building
```
Gerschtli, line 49: Maybe link to the open tracking issue in the description?

Gerschtli (commented):
Oh boy, what a dirty hack but yeah, as long as it works for you and it can be enabled and disabled with a flag, I'm fine with it.

Nice work to find and workaround this issue! Hopefully we can find the real cause behind this error soon.

overlays/htop.nix
```diff
@@ -7,12 +7,33 @@ let
 in
 
 {
-  htop = nixpkgs.htop.overrideAttrs (old: {
+  htop = nixpkgs.stdenv.mkDerivation rec {
+    pname = "htop";
+    version = "2.2.0";
```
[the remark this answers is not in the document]
t184256: I tried porting the patch to 3.0, but I couldn't make it work (it segfaulted). so this is the best I have now.

t184256:
Can't say I know what causes it, but shipping a workaround that's on by default is out of question: it'll be ugly and it might be just my device.
If there are more reports, then yeah, at least a Nix-side solution will be desired. If it's just me, it could be a nudge for me to use remote building.

I agree with the suggested changes, please take a second look.

Gerschtli:
Diff LGTM. Did not test it locally yet.
Travis is currently failing because of some cachix issues.

And for future PRs: You can use "fixup" commits and "autosquash". It improves the reviewing process because one can see the changes and still have a clean history :)

t184256:
Huh? I'm adding 'fixup' commits, then  moving them back in history and squashing them together with the comments introducing the change I amend. What do you suggest to do instead?

Gerschtli:
Sorry, I wasn't clear enough. Recently I learned about fixup commits in git, which make reviewing PRs easier while still having a clean history. I thought I should spread the knowledge about this feature because I haven't seen it being used as much as it should be. It's such a powerful feature :) See https://blog.sebastian-daschner.com/entries/git-commit-fixup-autosquash for a short explanation.

And to be clear: this current PR is totally fine, nothing left todo concerning my comments about fixup commits :)

t184256:
Thanks for the link. TIL, sounds like a couple of nice shortcuts/safeguards to my present workflow (I just did `commit -m fixup` and tried not to squash them with the wrong thing later during the rebase).

modules/environment/login/nix-on-droid.nix.default
```diff
@@ -28,15 +28,21 @@
...
   #   nix-channel --update
   # you can configure home-manager in here like
   #home-manager.config =
   #  { pkgs, ... }:
   #  {
+  #    # Read the changelog before changing this value
+  #    home.stateVersion = "20.09";
+  #
+  #    # Use the same overlays as the system packages
+  #    nixpkgs.overlays = config.nixpkgs.overlays;
```
Gerschtli, line 44: I think, I would prefer setting `home-manager.useGlobalPkgs = true;` which syncs the complete nixpkgs instance.

t184256: OK, I've dropped that one commit then (I don't like useGlobalPkgs enough to recommend it to everyone, too magical to my taste).

Gerschtli:
I found an issue with home-manager, when useUserPackages is enabled: `home.profileDirectory` points to `/etc/..` (see https://github.com/nix-community/home-manager/blob/master/modules/home-environment.nix#L414).

This causes an invalid file path (for hm-session-vars.sh) to be sourced on shell init.

We should create a PR/issue in home-manager to be able to set the path of `home.profileDirectory` correctly or add support for nix-on-droid natively like nixos and nix-darwin.

I can reproduce your issue with make, too. Do we know of any device where this issue doesn't occur? I will test your workaround but if we have a 100% rate of failing builds, that would be a show stopper for me. Did you find the commit in nixpkgs or any other change that introduced this issue?

t184256:
> I found an issue with home-manager, when useUserPackages is enabled: home.profileDirectory points to /etc/.. (see https://github.com/nix-community/home-manager/blob/master/modules/home-environment.nix#L414).

> This causes an invalid file path (for hm-session-vars.sh) to be sourced on shell init.

I don't get it. It's already enabled by default now, what should I look for? We have it on now, we used to test with it being on previously, why didn't that trip it?

> We should create a PR/issue in home-manager to be able to set the path of home.profileDirectory correctly or add support for nix-on-droid natively like nixos and nix-darwin.

No opinion here.

Gerschtli:
The issue is following line in `.profile`:
```bash
. "/etc/profiles/per-user/nix-on-droid/etc/profile.d/hm-session-vars.sh"
```

This file does not exist which is causing an error. The problem is that this specific line I linked earlier is optimized for usage with nixos and nix-darwin, where this modified path is needed. For nix-on-droid we need the standard value `~/.nix-profile`.

Your fix concerning the make issue worked for me, but as mentioned earlier I don't feel well releasing the current state as stable branch. What do you think?

t184256:
.profile: oh, ouch, I get it now, it's enabling bash in h-m that creates `.profile`; didn't have it. Yeah, we gotta override `profileDirectory` somehow.

make: Tough question, and the answer depends on what other options do we have, or, more specifically, whether we even have saner ones.

Gerschtli:
Yes, .profile and every other place, where profileDirectory is used. Unfortunately it is an internal option, which will make it difficult to override it outside of home-manager.

make: Currently I don't see any better options than your workaround. Don't know what we should do about it..

t184256:
.profile: feeling lazy today, see 24b45ff

make: still no clue, will play with remote building instead to see how good of an alternative it is.

modules/home-manager.nix
```diff
@@ -120,6 +122,10 @@ in
 
       environment.packages = mkIf cfg.useUserPackages cfg.config.home.packages;
 
+      environment.etc = if (! cfg.useUserPackages) then {} else {
+        "profiles/per-user/${config.user.userName}".source =
+          builtins.toPath "${config.user.home}/.nix-profile";
+      };
```
Gerschtli, line 128:
```suggestion
      environment.etc = mkIf cfg.useUserPackages {
        "profiles/per-user/${config.user.userName}".source =
          "${config.user.home}/.nix-profile";
      };
```

And maybe add a small comment, why this workaround is needed. But seems legit to me as a working fix.

t184256: Added a comment, changed to mkIf. Makes me worried about how many times I should've used mkIf in the past and hasn't.

t184256:
I've decided to release as is, cause I'm royally stuck with #91 and I don't see it being resolved shortly.

Gerschtli:
Nice, I'm fine with that!

PR #95, opened 2021-01-12
allow to override home-manager

bbigras:
I'm using flakes instead of channels.


Gerschtli:
`import <home-manager>` and `<home-manager>` is not the same. I'm not sure if this works as expected..

t184256:
Yep, `(import <home-manager>) + "/modules/modules.nix"` won't work.

bbigras:
Right. My bad. Thanks for taking a look.

Closing this in favor of #97

PR #96, opened 2021-01-12
use the same nixpkgs

bbigras:
I have no idea how to change the one at:

https://github.com/t184256/nix-on-droid/blob/2fba450df011a0316de2d6013784c02a79ff1e22/modules/nixpkgs.nix#L57-L59

https://github.com/t184256/nix-on-droid/blob/2fba450df011a0316de2d6013784c02a79ff1e22/modules/nixpkgs.nix#L165-L171

I get some infinite recursion if I change it to `_module.args.pkgs = pkgs;`

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -19,7 +19,7 @@ let
         _module.args.pkgs = mkDefault pkgs;
       }
       configModule
-    ] ++ import ./module-list.nix;
+    ] ++ import ./module-list.nix { inherit pkgs; };
   };
 
   failedAssertions = map (x: x.message) (filter (x: !x.assertion) rawModule.config.assertions);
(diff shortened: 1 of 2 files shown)


t184256:
By changing `<nixpkgs>` in `_pkgs = import <nixpkgs> (` to a different path?

Gerschtli:
Why would you need that in the first place? I would add a new option isUsedAsFlake, which when enabled disables the config block of nixpkgs.nix.

I would love to see a flake.nix added to this repo BTW :)

bbigras:
> By changing <nixpkgs> in _pkgs = import <nixpkgs> ( to a different path?

Thanks, I'll take a look with the flake version.

I'll make a flake version instead.

Closing this in favor of #97

PR #97, opened 2021-01-12
add flake.nix

bbigras:
Todo:

- [x] backwards-compatible
- [x] do flake.nix correctly
- [x] modules/nixpkgs.nix

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,21 +159,33 @@ you shouldn't need a binary cache for that.
 
 ## Nix flakes
 
-Example, to use with nix flakes:
+Note that nix flake support is experimental at the moment.
+Still, there's some minimal usage example.
+You can build an activation package by procuring flake-powered nix
+(`nix run nixpkgs.nixFlakes`,
+ `echo "experimental-features = nix-command flakes" >> ~/.config/nix/nix.conf`)
+writing a `flake.nix`:
 
 ```nix
 {
   description = "nix-on-droid configuration";
 
-  inputs.nix-on-droid.url = "/home/bbigras/src/nix-on-droid";
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/release-20.09";
+    nix-on-droid.url = "github:t184256/nix-on-droid/master";
+    nix-on-droid.inputs.nixpkgs.follows = "nixpkgs";
+  };
 
   outputs = { nix-on-droid, ... }: {
-    nix-on-droid = (nix-on-droid.lib.aarch64-linux.nix-on-droid { config = ./your_config.nix; } ).activationPackage;
+    nix-on-droid = (nix-on-droid.lib.aarch64-linux.nix-on-droid {
+      config = ./.config/nixpkgs/nix-on-droid.nix;
+    }).activationPackage;
   };
 }
 ```
 
-Build with `nix build .#nix-on-droid --impure`.
+building it with `nix build .#nix-on-droid --impure`
+and activating it with `result/activate`.
 
 ## Tips
 
--- a/README.md
+++ b/README.md
@@ -157,7 +157,23 @@ if you are planning to maintain a long-term fork that users can update from.
 In case you only care about updates through wiping the data,
 you shouldn't need a binary cache for that.
 
+## Nix flakes
 
+Example, to use with nix flakes:
+
+```nix
+{
+  description = "nix-on-droid configuration";
+
+  inputs.nix-on-droid.url = "/home/bbigras/src/nix-on-droid";
+
+  outputs = { nix-on-droid, ... }: {
+    nix-on-droid = (nix-on-droid.lib.aarch64-linux.nix-on-droid { config = ./your_config.nix; } ).activationPackage;
+  };
+}
+```
+
+Build with `nix build .#nix-on-droid --impure`.
 
 ## Tips
 
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,36 @@
+{
+  description = "Nix-enabled environment for your Android device";
+
+  inputs = {
+    flake-utils.url = "github:numtide/flake-utils";
+    home-manager = {
+      url = "github:nix-community/home-manager";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+  };
+
+  outputs = { self, nixpkgs, home-manager, flake-utils }:
+    flake-utils.lib.eachSystem [
+      "aarch64-linux"
+      "i686-linux"
+    ]
+      (system:
+        let pkgs = nixpkgs.legacyPackages.${system}; in
+        rec {
+          lib = {
+            nix-on-droid = { config }: import ./modules {
+              inherit pkgs config;
+              isFlake = true;
+              home-manager = (import home-manager { });
+            };
+          };
+
+          overlays = ./overlays;
+
+          apps.nix-on-droid = flake-utils.lib.mkApp {
+            drv = (pkgs.callPackage ./nix-on-droid { });
+          };
+          defaultApp = apps.nix-on-droid;
+        }
+      );
+}
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -1,6 +1,6 @@
 # Copyright (c) 2019-2020, see AUTHORS. Licensed under MIT License, see LICENSE.
 
-{ pkgs ? import <nixpkgs> { }, config ? null }:
+{ pkgs ? import <nixpkgs> { }, home-manager ? import <home-manager> { }, config ? null }:
 
 with pkgs.lib;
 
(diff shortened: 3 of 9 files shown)


Gerschtli:
One very important note to begin with: Everything should be backwards-compatible. Flakes are still experimental and we should not force anyone to use an unstable release of nix.

bbigras:
Gotcha. I was thinking we could use flake-compat[1] but it wouldn't work with channels.

1: https://nixos.wiki/wiki/Flakes#Using_flakes_project_from_a_legacy_Nix

bbigras:
Ready for review.

bbigras:
Should there be other system than "aarch64-linux"?

Note that to update `flake.nix`, you can run `nix flake update --impure --recreate-lock-file --commit-lock-file` and it will produce a nice commit. The lock file could be excluded too I guess (home-manager doesn't have it in their repo).

flake.lock
```diff
@@ -0,0 +1,63 @@
+{
```
Gerschtli, line 1: The flake.lock should not be commited in this repo, because it is a library (see home-manager, nixpkgs).

flake.nix
```diff
@@ -0,0 +1,33 @@
...
+    flake-utils.url = "github:numtide/flake-utils";
+    home-manager = {
+      url = "github:rycee/home-manager";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+  };
+
+  outputs = { self, nixpkgs, home-manager, flake-utils }:
+    flake-utils.lib.eachSystem [
+      "aarch64-linux"
```
Gerschtli, line 15: i686 is an official supported system, too.

flake.nix
```diff
@@ -0,0 +1,33 @@
+{
+  description = "Nix-enabled environment for your Android device";
+
+  inputs = {
+    nixpkgs.url = "nixpkgs/nixos-unstable";
```
Gerschtli, line 5: You can omit this line. nixpkgs is a default input and the channel is irrelevant here

flake.nix
```diff
@@ -0,0 +1,33 @@
...
+    flake-utils.url = "github:numtide/flake-utils";
+    home-manager = {
+      url = "github:rycee/home-manager";
```
Gerschtli, line 8:
```suggestion
      url = "github:nix-community/home-manager";
```

flake.nix
```diff
@@ -0,0 +1,33 @@
...
+        rec {
+          lib = {
+            nix-on-droid = config: import ./modules {
+              inherit pkgs config;
+              home-manager = (import home-manager { });
+            };
+          };
+
+          apps.nix-on-droid = flake-utils.lib.mkApp {
+            drv = (import ./default.nix { inherit pkgs; });
```
Gerschtli, line 28: Could you call nix-on-droid directly without importing default.nix first? default.nix is only needed for channels, would love to keep this.

modules/default.nix
```diff
@@ -16,10 +16,12 @@ let
   rawModule = evalModules {
     modules = [
       {
+        _module.args.home-manager = home-manager;
         _module.args.pkgs = mkDefault pkgs;
+        _module.args.pkgsPath = pkgs.path;
```
Gerschtli, line 21: You don't need this, just use pkgs.path whereever you need it or do I miss something?

modules/nixpkgs.nix
```diff
@@ -54,7 +54,7 @@ let
     merge = lib.mergeOneOption;
   };
 
-  _pkgs = import <nixpkgs> (
+  _pkgs = import pkgsPath (
```
Gerschtli, line 57: As mentioned in another PR of yours, this module should not be used when running it as flake IMO. Seems to be a bit impure to me. But I'm open for a discussion.

flake.nix
```diff
@@ -0,0 +1,33 @@
...
+          };
+          defaultApp = apps.nix-on-droid;
+        }
```
Gerschtli, line 31: You should add the overlays here (`./overlays`)

Gerschtli (changes requested):

bbigras:
Thanks for the review. I think I made all the changes.

Gerschtli (approved):
Not tested locally but the diff LGTM. Maybe you could split the one big commit into smaller ones, but that just a matter of taste.

bbigras:
Would 2 commits be enough (1 for the optional args and 1 for the flake stuff)?

Gerschtli:
Yeah, sounds good.

modules/module-list.nix
```diff
@@ -10,10 +12,9 @@
...
   ./home-manager.nix
-  ./nixpkgs.nix
   ./time.nix
   ./user.nix
   ./version.nix
   ./workaround-make.nix
-  <nixpkgs/nixos/modules/misc/assertions.nix>
-]
+  (pkgs.path + "/nixos/modules/misc/assertions.nix")
+] ++ pkgs.lib.optionals (!isFlake) [ ./nixpkgs.nix ]
```
Gerschtli, line 20:
```suggestion
] ++ pkgs.lib.optional (!isFlake) ./nixpkgs.nix
```

bbigras: Yeah I used `optionals` because of https://discourse.nixos.org/t/lib-optional-considered-harmful/7081 but I'll make the change.

Gerschtli: Oh good point, you can leave it like that :)

bbigras:
I just pushed the changes in 2 commits.

Gerschtli:
Nice, thank you. One thing I thought about is changing the signature of lib.nix-on-droid to accept a set with configuration as a key. This would simplify future changes while maintaining backwards-compatibility.

Maybe an example in the readme how to use it and a mention in the changelog would be nice.

Unfortunately I don't have time to test this, maybe @t182256 has.

bbigras:
> Nice, thank you. One thing I thought about is changing the signature of lib.nix-on-droid to accept a set with configuration as a key. This would simplify future changes while maintaining backwards-compatibility.

Like this?

t184256:
> Unfortunately I don't have time to test this, maybe @t184256 has.

I'll have some time this week, but I haven't been paying proper attention so far,
so I'd appreciate a brief test plan.
What should I test, besides the usual clean install and upgrade?
How far are we from a fully flake-based config?

bbigras:
> Maybe an example in the readme how to use it and a mention in the changelog would be nice.

Done. I did a very basic example. Feel free to suggest changes.

> How far are we from a fully flake-based config?

Right now I can deploy from my flake-based [nix-config](https://github.com/bbigras/nix-config/) to my phone using `deploy .#pixel2 -- --impure`.

see:
- https://github.com/bbigras/nix-config/blob/d002fafe0c9a8d665a2dad761e97566203f170f4/nix/deploy.nix#L78-L91
- https://github.com/bbigras/nix-config/blob/d002fafe0c9a8d665a2dad761e97566203f170f4/hosts/pixel2/default.nix

Note that my github actions can't build for aarch64.

Also note that my example doesn't build without `--impure`:
`builtins.storePath' is not allowed in pure evaluation mode`

Gerschtli:
First of all, this change should not break anything. Further than that, because the nix-on-droid script does not work with flakes yet, I would just try to build a working generation on nix-on-droid.

We should mark the flakes integration experimental and subject to change because not everything is setup yet to use nix-on-droid with flakes.

I think it would be interesting to discuss what would be necessary to bootstrap nix-on-droid with nix flake support only, so you could choose between channels and flakes on bootstrap for example.

One question: why is --impure needed and what do we need to do for it to be no longer necessary?

t184256:
I'm for merging the changes, as they don't break anything. Give me some more time and I'll tweak the README example a bit, so that it's something replicateable by anybody w/o remote building.

Also, just a thought: a simple, though hackish way would be to pick up uid/gid from the configuration file, and only go the `idDerivation` way if it's not there. This way "external" building won't have to be parametrized externally.

t184256:
All right, I've added three small commits on top of that, please review.

@bbigras, thanks for your work!

@Gerschtli, you can probably merge it right away if you find my changes fine.

README.md
```diff
@@ -157,7 +157,35 @@ if you are planning to maintain a long-term fork that users can update from.
 In case you only care about updates through wiping the data,
 you shouldn't need a binary cache for that.
 
+## Nix flakes
 
+Note that nix flake support is experimental at the moment.
+Still, there's some minimal usage example.
+You can build an activation package by procuring flake-powered nix
+(`nix run nixpkgs.nixFlakes`,
+ `echo "experimental-features = nix-command flakes" >> ~/.config/nix/nix.conf)
```
Gerschtli, line 166: A closing back tick is missing

README.md
```diff
@@ -157,7 +157,35 @@ if you are planning to maintain a long-term fork that users can update from.
...
+  outputs = { nix-on-droid, ... }: {
+    nix-on-droid = (nix-on-droid.lib.aarch64-linux.nix-on-droid {
+      config = ./.config/nixpkgs/nix-on-droid.nix;
+    }).activationPackage;
+  };
+}
+```
+
+building it with `nix build .#nix-on-droid --impure`
+and activating it with `result/activate`.
```
Gerschtli, line 188: I am not sure if running this activate script off the result dir will work nicely with the nix-env --set, we are using to manage our generstions. But because of the experimental state, I can live with that line.

Gerschtli (approved):
LGTM, can merge it, after the missing back tick got added in the readme.

README.md
t184256: Added, thanks for noticing.

PR #100, opened 2021-01-19
user: allow to set uid/gid

bbigras:
> Also, just a thought: a simple, though hackish way would be to pick up uid/gid from the configuration file, and only go the idDerivation way if it's not there. This way "external" building won't have to be parametrized externally.

https://github.com/t184256/nix-on-droid/pull/97#issuecomment-761821403

@t184256 would this simple PR work? I didn't do any "and only go the idDerivation way if it's not there" thing.

### DIFF
--- a/modules/user.nix
+++ b/modules/user.nix
@@ -10,8 +10,8 @@ let
   idsDerivation = pkgs.runCommandLocal "ids.nix" {} ''
     cat > $out <<EOF
     {
-      gid = "$(${pkgs.coreutils}/bin/id -g)";
-      uid = "$(${pkgs.coreutils}/bin/id -u)";
+      gid = "${toString cfg.gid}";
+      uid = "${toString cfg.uid}";
     }
     EOF
   '';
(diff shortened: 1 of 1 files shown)


Gerschtli:
This won't work because these values need to be evaluateted by bash and cannot be casted to an int, these are just strings for nix

bbigras:
It seems to work in my case but do you mean it won't work for the default case?

Gerschtli:
Yes, because `toInt "$(${pkgs.coreutils}/bin/id -u)"` does not evaluate to the needed uid value.

t184256:
Yeah, this won't work without the device, because you need to install the app to know the uid/gid.

What I meant was baking these into the config during the initial installation, maybe in a form of a comment, so that remote builders can just uncomment it, but local builders knew it's optional. Not sure if it's worth the hassle though, but, on the other hand, it shouldn't hurt too.

bbigras:
Thanks!

Issue #69, opened 2020-07-11
root certificates missing?


bbigras:
I got this message when I try to use `niv`:

![image](https://user-images.githubusercontent.com/24027/87212591-53f7ae00-c2ed-11ea-81ef-51518a482780.png)

I have `cacert` installed.

I also have a similar message when I try to push stuff to cachix.

Any ideas?

t184256:
Wild guess: https://knowledge.digicert.com/alerts/DigiCert-ICA-Replacement

ahdyt:
Hi I have workaround for this issue

First install curl

Then mkdir at /etc/ssl/certs

cd there and run

```curl --remote-name --time-cond cacert.pem https://curl.se/ca/cacert.pem```

Perhaps there's one who like to make automated script, that would be useful.

**Comment by Gerschtli** [comment]
A more robust solution would be to add
```nix
{
  environment.etc."ssl/certs/ca-certificates.crt".source = "${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt";
}
```
Or a modified version of https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/security/ca.nix.

Maybe anyone has the time to add such a module to nix-on-droid to fix it for all users :)

ahdyt:
> Maybe anyone has the time to add such a module to nix-on-droid to fix it for all users :)

the module still foreign for me

PR #115, opened 2021-05-27
ca: add certificates in /etc/ssl/certs

Gerschtli:
Fixes https://github.com/t184256/nix-on-droid/issues/69


t184256 (approved):
Works as expected.

Issue #94, opened 2021-01-09
Any way to deploy from desktop to phone using deploy-rs or nixus?


bbigras:
https://github.com/serokell/deploy-rs
https://github.com/Infinisil/nixus

Those tools allow us to build multiples machines at once and push deploy them.

If I can build aarch64, can I just build something, nix-copy it to the phone and run an `activate` command on the phone?

Gerschtli:
That should be possible I think. Have a look at `nix-on-droid/nix-on-droid.sh` on how we build and activate :)

t184256:
You'll need an ssh server on the phone if you're keen on initiating that from a desktop, I guess, otherwise it's just https://github.com/t184256/nix-on-droid/wiki/Remote-building.

But once you have it, "build something, nix-copy it to the phone and run an activate command on the phone" should just work, yeah.

bbigras:
Thanks. I think I'm making some progress.

Any way to force `system` to `aarch64-linux` when running on `x86_x64`?

bbigras:
Current status:

```
❯ deploy ".#pixel2" -- --impure
🚀 ℹ️ [deploy] [INFO] Running checks for flake in .
warning: Git tree '/home/bbigras/nix-config' is dirty
warning: unknown flake output 'deploy'
🚀 ℹ️ [deploy] [INFO] Evaluating flake in .
warning: Git tree '/home/bbigras/nix-config' is dirty
🚀 ℹ️ [deploy] [INFO] The following profiles are going to be deployed:
[pixel2.system]
user = "nix-on-droid"
ssh_user = "nix-on-droid"
path = "/nix/store/zfyia9axhilazg3l1qw8vhi2n2raqj2d-activatable-nix-on-droid-generation"
hostname = "pixel2"
ssh_opts = []

🚀 ℹ️ [deploy] [INFO] Building profile `system` for node `pixel2`
warning: Git tree '/home/bbigras/nix-config' is dirty
copying path '/nix/store/2jjf28m0qvn9cfqrii9gzav5msv0nf0r-bash-interactive-4.4-p23-man' from 'https://cache.nixos.org'...
copying path '/nix/store/q3acm62dxjdw7rfgnz4zyiz2j619ngr3-glibc-locales-2.32-10' from 'https://cache.nixos.org'...
copying path '/nix/store/qq58n27vwpbhqbzzrzwq0gvl64nl7v74-less-563' from 'https://cache.nixos.org'...
[...]

PR #119, opened 2021-06-21
flake.nix: fix overlays

bbigras:
With this change, the overlays are used when using flakes.

related to https://github.com/t184256/nix-on-droid/issues/94#issuecomment-850849580

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -10,12 +10,18 @@
   };
 
   outputs = { self, nixpkgs, home-manager, flake-utils }:
+    {
+      overlays = import ./overlays;
+    } //
     flake-utils.lib.eachSystem [
       "aarch64-linux"
       "i686-linux"
     ]
       (system:
-        let pkgs = nixpkgs.legacyPackages.${system}; in
+        let pkgs = import nixpkgs {
+          inherit system;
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -25,8 +31,6 @@
             };
           };
 
-          overlays = ./overlays;
```
Gerschtli, line 28: Just to be sure: This nested overlays field is not part of the standard flake specification, right?

Gerschtli (approved):
LGTM, thank you!

Note to myself: Someday I have to port my config to flakes.. It's just so awesome :D

flake.nix
t184256: Yeah, as far as I can see it should be toplevel in the output schema, and not branched with `flake-utils.lib.eachSystem`.

t184256:
Sorry it took so long, I'm not using flakes yet on my nix-on-droid device.

Looks good, applies overlays.

Thanks @bbigras and also thanks @notgne2.

flake.nix
bbigras: I have no idea, but I thought the overlays didn't need to be generated for each system.

PR #122, opened 2021-06-24
21.05

t184256:
@Gerschtli, got time for a review? Nothing fancy going on in here.

Also, got any more suggestions for the changelog?


Gerschtli:
Diff LGTM, didn't grep for 20.09 by myself but looks like you found every occurence.

Did not test it on my device, will check it on the weekend.

Gerschtli:
Please do not forget to push all packages and the ones affected by overlays to push to the binary cache :)

modules/workaround-make.nix
```diff
@@ -59,6 +59,11 @@ in
   ###### implementation
 
   config = {
+    warnings = lib.optionals cfg.enable [
+      ("system.workaround.make-posix-spawn.enable " +
+       "should no longer be needed on 21.05")
```
Gerschtli, line 64: Maybe clarify that this option will be removed in 21.11 and if anyone still needs this, they should create an issue?

Gerschtli (approved):
Tested it and could not find any issues :)

modules/home-manager.nix
```diff
@@ -72,7 +72,7 @@ in
         installation of user packages through the
         <option>environment.packages</option> option.
       '' // {
-        default = versionAtLeast config.system.stateVersion "20.09";
+        default = versionAtLeast config.system.stateVersion "21.05";
       };
```
Gerschtli, line 76: This should not be changed!

t184256: uh-oh, fixup time

modules/workaround-make.nix
t184256: expanded the warning text

modules/home-manager.nix
t184256: changed back, thanks for catching it!

t184256:
OK, let's release. Thank you for your review!

PR #124, opened 2021-06-29
overlays/lib/nixpkgs.nix: use system

bbigras:
Any idea what to do with this:

https://github.com/t184256/nix-on-droid/blob/78d349eecc2382b65c6d93f4ef15d5588db169c6/modules/nixpkgs.nix#L169

Does non-flake use `system`?

EDIT: maybe `{ system ? builtins.currentSystem }:`

### DIFF
--- a/overlays/lib/nixpkgs.nix
+++ b/overlays/lib/nixpkgs.nix
@@ -13,6 +13,6 @@ let
 in
 
 import pinnedPkgsSrc {
-  inherit (super) config;
+  inherit (super) config system;
   overlays = [ ];
 }


bbigras:
Can `arch` be used for `system`?

`system` seems to use the `i686-linux` format while `arch` is `i686`.

Gerschtli:
Why not just using `super.system` instead of injecting it?

bbigras:
> Why not just using `super.system` instead of injecting it?

I didn't know about that.

I updated the commit.

Gerschtli:
LGTM, can not test it because I don't have a setup with flakes.

bbigras:
Thanks!

Issue #128, opened 2021-07-18
Provide something like nixos `nix.package` option


NobbZ:
Currently if one wants to use an unstable version of nix they have to manually `nix-shell` into it after each start. Just adding it to `environment.packages` doesn't seem to properly persist.

With `environment.packages = [ pkgs.nixUnstable ]` `nix --version` still reports 2.3.x. even after restarting the terminal.

I have to explicitly do `nix-shell -p nixUnstable` first.

nixOS Module system has [`nix.package`](https://search.nixos.org/options?channel=21.05&show=nix.package&from=0&size=50&sort=relevance&query=nix.package) for this.

Gerschtli:
Please have a look at https://github.com/t184256/nix-on-droid/pull/129 :)

PR #129, opened 2021-07-24
nix: add package option

Gerschtli:
Resolves https://github.com/t184256/nix-on-droid/issues/128

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Release 21.11 (unreleased)
+
+### New Options
+
+* The `nix.package` can be used to set the system-wide nix package.
+
 ## Release 21.05
 
 ### New Options
--- a/modules/build/activation.nix
+++ b/modules/build/activation.nix
@@ -1,4 +1,4 @@
-# Copyright (c) 2019-2020, see AUTHORS. Licensed under MIT License, see LICENSE.
+# Copyright (c) 2019-2021, see AUTHORS. Licensed under MIT License, see LICENSE.
 
 { config, lib, pkgs, ... }:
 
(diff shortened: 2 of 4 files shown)


t184256:
Hi, thanks for looking into this.

1) could you bump copyright years on the affected files? might be silly, but still
2) `nix-on-droid` script won't pick the selected version, right? at least I don't see how it would.

Gerschtli:
Will bump the years in a minute.
nix-on-droid should use the defined nix through https://github.com/t184256/nix-on-droid/pull/129/files#diff-dd4183ac01238a8af6cb39b8e12afd4fc614d11cf15167ce7a9c8712d0c337d3R45

t184256:
How could I possibly overlook that, I wonder =o

t184256:
What I'm up to: playing with it by setting it to `nixFlakes` and trying to evict the old nix from the system entirely. Went down to the single referrer being the generation itself (why, how), but now am unwinding and recovering back from `nix-on-droid` not liking `nixFlakes` as its `nix`.

Gerschtli:
So there is still a reference to the default nix package despite changing this option? Can you clarify where the reference is exactly?

t184256:
If only I knew how to check.

Assorted notes so that it's easier for me to retrace:

1) just setting it to `nixFlakes` isn't enough, one also needs at least `experimental-features = nix-command`
2) after rebuilding a generation with `nixFlakes` and re-logging in, it loses the `channels` part of `$NIX_PATH`, rendering `nix-on-droid` unusable until one corrects that as well
3) `nix-store --query --referrers` lists generation and, oddly enough, channels. channels go away after `nix-channel update` with a new nix, hasn't figured out how to understand why derivation refers to old nix (dump it as a nar, maybe?)

t184256:
Found it! It's in `activation` itself, `modules/build/activation.nix:22`.

Gerschtli:
Nice thank you

t184256:
OK, I haven't dug to the bottom of why `$NIX_PATH` is the way it is (and stops being the way it is with `nixFlakes`), but this absolutely doesn't devalue this PR, so, merging. Thank you!

t184256 (approved):

PR #132, opened 2021-09-10
Backport to 21.05 option `nix.package`

ShamrockLee:
Not sure if we do backports here, but it doesn't hurt.

**Motivation**

1.  NixOS/nixpkgs/release-21.05 is the current stable branch.
2.  The backport is backward-compatible. A similar option has been part of the NixOS configuration long before they are added here.
3.  There's no workaround to specify the Nix package to use without adding this option. This patch greatly increase the user experience for flake-based configuration (`nix.package = pkgs.nixUnstable;`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 * The `/etc/nix/nix.conf` file is now fully configurable with the
   new options `nix.*`.
 
+* [Backport from 21.11] The `nix.package` can be used to set the system-wide nix package.
+
 ### Deprecations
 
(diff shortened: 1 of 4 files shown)


t184256:
I've thought about that for a while, and I don't think backporting is a good time investment given that our master branch is always or almost always compatible with previous nixpkgs. Unless somebody steps up to test both backports and non-ff upgrades to new releases, I'd prefer not to do backports.

ShamrockLee:
@t184256 I had thought that there would be some version checks the way the home-manager does.
Backporting would not be necessary if stable nixpkgs users can use the master nix-on-droid directly.

Issue #111, opened 2021-05-09
Confusion on the usage of Flakes


Bobbbay:
Hey y'all, I have a feeling this is a stupid question...

So, I activated Flakes. Well, actually, it errored out:

```
error: selector 'result' matches no derivations
```

But it seems Flakes have been activated? I can run `nix flake`, anyhow.

Now, I'm slightly confused as to where to continue from there. On my NixOS computers, I run `nixos-rebuild switch` and we're good from there. However, `nix-on-droid switch` still seems to be using a non-flaked config.

Super confused as to my next steps :D

t184256:
Figure it out and tell us =) Flakes are not supported yet; patches are welcome.

ShamrockLee:
I can confirm this issue.
Things seems work except the confusing error following the long process time trying to `Activate linkProfile`.

ShamrockLee:
The problem seems to lie in the activation script instead of the `nix-on-droid` command.

It seems that `activationPackage` is an attribute of the output of [modules/default.nix](https://github.com/t184256/nix-on-droid/tree/master/modules/default.nix), and the Nix expressions related to the `activationPackage` generation is mainly inside [modules/build/activation.nix](https://github.com/t184256/nix-on-droid/tree/master/modules/build/activation.nix).

Gerschtli:
That is not entirely true. The main reason for this switch is to allow rollbacks to reuse the already built generation. The problem when your building your nix-on-droid build may be that generationDir is the path to your "result" symlink instead of the /nix/store path. Maybe you can try resolving the symlink of generationDir in the else case.

ShamrockLee:
@t184256 @Gerschtli 

Sorry for not well-testing the commit.

For some strange reason, the path given by `$(dirname $0)` from the activation script `result/activate` seems not to exist for the tooling inside the script. My above PR (#135) needs to be reverted.

t184256:
How does that happen? And how would a revert help?

ShamrockLee:
@t184256 I'm not sure if the revert help, since the problem is already there berofe the commit What I'm sure is that my above commit doesn't solve the problem.

t184256:
Then reason is https://github.com/t184256/nix-on-droid/blob/3aa377bf01ce82da48567d026ab9f2ad3f1befc2/modules/build/activation.nix#L38. Drop it from the activation script, run it from non-$HOME and see what breaks?

ShamrockLee:
[...]

PR #135, opened 2021-09-12
activation.nix: fix for "activation from build result"

ShamrockLee:
Set profile with the store path instead of the path of the "result" link.

Solve #111


t184256 (approved):
No objections to that change, thank you.

PR #136, opened 2021-09-20
activation.nix: fix for "activation from build result" (revised)

ShamrockLee:
Solve #111 

Fix the activation script by dropping the line

```bash
cd $HOME
```

in `activation.nix`

### DIFF
--- a/modules/build/activation.nix
+++ b/modules/build/activation.nix
@@ -35,9 +35,9 @@ let
     set -eu
     set -o pipefail
 
-    cd $HOME
-
     export PATH="${activationBinPaths}"
+    _NOD_GENERATION_DIR="$(realpath "$(dirname "$0")")"
+    cd "$HOME"
 
     ${builtins.readFile ../lib-bash/color-echo.sh}
     ${builtins.readFile ../lib-bash/activation-init.sh}
@@ -133,7 +133,7 @@ in
         if [[ $generationDir =~ ^${profileDirectory}-([0-9]+)-link$ ]]; then
           $DRY_RUN_CMD nix-env --profile "${profileDirectory}" --switch-generation "''${BASH_REMATCH[1]}"
         else
-          $DRY_RUN_CMD nix-env --profile "${profileDirectory}" --set "$(realpath "$generationDir")"
+          $DRY_RUN_CMD nix-env --profile "${profileDirectory}" --set "$_NOD_GENERATION_DIR"
         fi
       '';
 


Gerschtli:
The problem is that the script loses some predictability. Ideally, nobody should ever rely on the current working directory but to make it completely reproducable, I do not feel right removing that. Not sure if it just hides the real cause..

t184256:
I'd say the script loses a guard against impredictability, and as long as everything (else than the `$0`) deals in absolute paths it should be OK.

What better options do we have, find a way to inject the absolute path during build-time?

t184256:
How I understand it:

1. one builds a configuration with flakes at `/some/path`, obtaining `/some/path/result/activate`.

2. they try activating it with `./result/activate`, `$0` resolves to `./result/activate`, `$generationDir` is going to be `./result`.

3. As the script `cd`s into `$HOME`, forgetting the former cwd, it loses the info needed to resolve the path to pass to `--set`.

Options:

1. Do not `cd` into `$HOME`, downside: fragility
2. Resolve `$0` to the right path first, then `cd` to a predictable path
3. Bake in the path during build time, if possible; IDK if possible
4. ???

Gerschtli:
Ah got it now, thank you. My approach would be to save the completely resolved (follow symlinks, resolve relative paths) script name (`$0`) in a variable the line before calling `cd $HOME`. Then just use the variable.

So option 2.

modules/build/activation.nix
```diff
@@ -35,8 +35,6 @@ let
     set -eu
     set -o pipefail
```
Gerschtli, line 37: Something like that
```suggestion

    GENERATION_DIR="$(realpath "$(dirname "$0")")"
    cd "$HOME"
```

ShamrockLee:
The only CWD-dependent part I found in the activation script is the "$0" line. What is the purpose of `cd "$HOME"` then?

modules/build/activation.nix
ShamrockLee: Applied.

Gerschtli (approved):
Tested it locally and works like a charm.

@t184256 Would love to see that merged :)

Gerschtli:
@ShamrockLee The purpose of `cd "$HOME"` is to minimize possible difference introduced by how the script has been called. Ideally, nobody relies on the CWD, but the more predictive the better :)

NobbZ:
> Ideally, nobody relies on the CWD, but the more predictive the better :)

Ideally, I can rely on the CWD, to be able to narrow down the scope of things to consider.

Gerschtli:
Not sure what you are trying to say, but I agree that the CWD should be reliably set. What I meant was that ideally everybody only works with absolute paths instead of relative ones to reduce potential errors.

NobbZ:
If I were passing `--flake $(pwd)` instead, would it have worked? This is inconvinient. Most of the world works using relative pathes, and accepting the CWD as a helpfull state. Even nix does.

Gerschtli:
Do you have concrete problem with the current implementation? I do not get the issue you are trying to solve.

NobbZ:
I do not have any problems with the implementation, I have a problem with the sentiment that relying on "CWD" was a bad thing.

Gerschtli:
I will give you an example why relying on the CWD in the activation script (the script we are talking about) is potentially problematic:

Anyone can write parts of the activation script through setting the options `build.activation*`. If you rely or change the CWD in there and do not return to the initial CWD and any following parts of the script relies on the CWD being e.g. `$HOME` which it is not anymore, these parts of the activation script (in the best case) fail or (in the worst case) breaking some stuff. If you want to rely on the CWD, always make sure to ensure, you are at the CWD you expect and go back to the CWD set before when your part is done. Because this is error-prone, I suggest to not rely on the CWD in the activation script.

NobbZ:
If you see it like that, you might want to provide `-o` to the underlying `nix build`. Then there is no need anymore to rely on any relative pathes.

Gerschtli:
In the activation script, nix-build is never called. Can you show me the line you are talking about?

NobbZ:
I'm not talking about the activation script. I'm talking about the protocol between the build and the activation.

If both would agree on a well known specified absolute location, there was no need to hold this discussion at all. Though if the location is not well specified and might happen to be relative to the PWD, then there is no workaround but to adhere to the PWD.

Gerschtli:
Then this discussion is obsolete because we are talking only about the activation script. The "protocol between" does not have problems with CWDs AFAIK.

PR #138, opened 2021-10-18
Bind /dev/shm

azuwis:
POSIX semaphores need /dev/shm to work. Without this, python
multiprocessing gives errors:

```
Traceback (most recent call last):
  ...
  File "/usr/lib/python2.7/multiprocessing/synchronize.py", line 75, in __init__
    sl = self._semlock = _multiprocessing.SemLock(kind, value, maxvalue)
OSError: [Errno 13] Permission denied
```

### DIFF
--- a/modules/environment/login/login.nix
+++ b/modules/environment/login/login.nix
@@ -33,6 +33,7 @@ writeScript "login" ''
     -b ${installationDir}/bin:/bin \
     -b ${installationDir}/etc:/etc \
     -b ${installationDir}/tmp:/tmp \
+    -b ${installationDir}/tmp:/dev/shm \
     -b ${installationDir}/usr:/usr \
     -b /:/android \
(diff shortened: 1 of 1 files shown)


modules/environment/login/login.nix
```diff
@@ -33,6 +33,7 @@ writeScript "login" ''
     -b ${installationDir}/bin:/bin \
     -b ${installationDir}/etc:/etc \
     -b ${installationDir}/tmp:/tmp \
+    -b ${installationDir}/tmp:/dev/shm \
```
Gerschtli, line 36:
```suggestion
    -b ${installationDir}/dev/shm:/dev/shm \
```

Gerschtli (commented):
At least on my phone (not rooted) I do not have access to `/dev/shm`. I am not sure if there is a workaround for that..

PR #140, opened 2021-10-22
Remove set +u/-u

azuwis:
Lots of zsh plugins fail to load when login via sshd:

```
$ ssh -p 8022 nix-on-droid@192.168.0.1
(eval):28: __zoxide_hooked: parameter not set
(eval):8: precmd_functions[(r)_direnv_hook]: parameter not set
async_init:1: ASYNC_INIT_DONE: parameter not set
prompt_pure_setup:10: prompt_newline: parameter not set
/nix/store/drwh7ngjshx794jvrvfvaz7wacjhkywi-zsh-fast-syntax-highlighting-1.54/share/zsh/site-functions/fast-syntax-highlighting.plugin.zsh:48: ZPLG_CUR_PLUGIN: parameter not set
```

### DIFF
--- a/modules/environment/login/login-inner.nix
+++ b/modules/environment/login/login-inner.nix
@@ -60,9 +60,7 @@ writeText "login-inner" ''
     fi
   ''}
 
-  set +u
   . "${config.user.home}/.nix-profile/etc/profile.d/nix-on-droid-session-init.sh"
-  set -u
 
   ${lib.optionalString config.build.initialBuild ''
     exec /usr/bin/env bash  # otherwise it'll be a limited bash that came with Nix
--- a/modules/environment/session-init.nix
+++ b/modules/environment/session-init.nix
@@ -24,9 +24,7 @@ let
       ${optionalString (config.home-manager.config != null) ''
         if [ -e "$HOME/.nix-profile/etc/profile.d/hm-session-vars.sh" ]; then
           export NIX_PATH=$HOME/.nix-defexpr/channels''${NIX_PATH:+:}$NIX_PATH
-          set +u
           . "$HOME/.nix-profile/etc/profile.d/hm-session-vars.sh"
-          set -u
         fi
       ''}
 


Gerschtli:
Any further explanation from you, @azuwis? As long as we don't understand your proposed fix and there is no activity here, I would propose to close this PR.

It can be reopened anytime.

azuwis:
Zsh in sshd gives errors without this change, but strangely zsh in termux does not.

I'll try to find out the real cause of the errors, and provide a proper fix.

Gerschtli:
I would say the whole ssh setup needs some work for me, I always need to run zsh manually after login to have my environment correctly setup, maybe this is related to your issue. I am looking forward to this being fixed :)

azuwis:
Currently, important env vars for termux shell to work, like `PATH` are initialized in `/data/data/com.termux.nix/files/usr/usr/lib/login-inner`:

```
# This file is generated by nix-on-droid. DO NOT EDIT.

set -eu -o pipefail

if [ "$#" -eq 0 ]; then  # if script is called from within nix-on-droid app
  echo "Welcome to Nix-on-Droid!"
  echo "If nothing works, open an issue at https://github.com/t184256/nix-on-droid/issues or try the rescue shell."
fi

set +u
. "/data/data/com.termux.nix/files/home/.nix-profile/etc/profile.d/nix-on-droid-session-init.sh"
set -u

usershell="/nix/store/y8y55zpyqqiz86qsrjrfl7nz2h7s0gbl-zsh-5.8/bin/zsh"
if [ "$#" -gt 0 ]; then  # if script is not called from within nix-on-droid app
  exec /usr/bin/env "$@"
elif [ -x "$usershell" ]; then
  exec -a "-${usershell##*/}" "$usershell"
else
  echo "Cannot execute shell '/nix/store/y8y55zpyqqiz86qsrjrfl7nz2h7s0gbl-zsh-5.8/bin/zsh', falling back to bash"
  exec -l bash
fi
```

But AFAIK, sshd will clear most of this env vars before invoke shell, that's why shell in sshd does not work out of box.

For zsh in sshd to work, I use home-manager's `programs.zsh.envExtra`:

```
programs.zsh.envExtra = ''
. "/data/data/com.termux.nix/files/home/.nix-profile/etc/profile.d/nix-on-droid-session-init.sh"
'';
```

Which will generate `~/.zshenv`.

The content of `/data/data/com.termux.nix/files/home/.nix-profile/etc/profile.d/nix-on-droid-session-init.sh` is:

```
# Only source this once.
[ -n "$__NOD_SESS_INIT_SOURCED" ] && return
export __NOD_SESS_INIT_SOURCED=1

. $HOME/.nix-profile/etc/profile.d/nix.sh

# workaround for nix 2.4, see https://github.com/NixOS/nixpkgs/issues/149791
export NIX_PATH=$HOME/.nix-defexpr/channels${NIX_PATH:+:}$NIX_PATH

if [ -e "$HOME/.nix-profile/etc/profile.d/hm-session-vars.sh" ]; then
  set +u
  . "$HOME/.nix-profile/etc/profile.d/hm-session-vars.sh"
  set -u
fi

# Workaround for https://github.com/NixOS/nix/issues/1865
export NIX_PATH=nixpkgs=$HOME/.nix-defexpr/channels/nixpkgs/:$NIX_PATH

export GC_NPROCS="1"
export HOME="/data/data/com.termux.nix/files/home"
export LOCALE_ARCHIVE="/nix/store/gxzddl950k06f4r7vd11ib1wn7ki6h1m-glibc-locales-2.33-56/lib/locale/locale-archive"
export TZDIR="/etc/zoneinfo"
export USER="nix-on-droid"
```

So when zsh load `~/.zshenv`, `set -u` got run. Zsh then loads `~/.zshrc`, and `set -u` causes error message in this issue.

azuwis:
I think `nix-on-droid-session-init.sh` should be put in shell rc files, like `/etc/bashrc`, `/etc/zshenv`, so termux shell and sshd shell will behave the same way.

Gerschtli:
Thank you for the investigation.

> I think nix-on-droid-session-init.sh should be put in shell rc files, like /etc/bashrc, /etc/zshenv, so termux shell and sshd shell will behave the same way.

That sounds reasonable, I will have a look at that.

Gerschtli:
@azuwis Couldn't figure out how to push to the PR so please see https://github.com/t184256/nix-on-droid/pull/153 for my proposed fix. This PR should be closed then. I cherry-picked your commit into my PR :)

PR #141, opened 2021-11-19
Expose options

lucasew:
Options are useful for documentation purposes and can be used by things like [nix-option](https://github.com/lucasew/nix-option) and automatic documentation generation systems.


t184256 (approved):
I don't see why not =)

PR #145, opened 2021-12-03
flake.nix: Clean up, allow overriding pkgs

zhaofengli:
This allows `pkgs` and `home-manager` to be easily overridden for custom configs. We also fix the `overlay` output so that the output attribute set matches the standard schema (`nix flake check`).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -9,32 +9,24 @@
     };
   };
 
-  outputs = { self, nixpkgs, home-manager, flake-utils }:
-    {
-      overlays = import ./overlays;
-    } //
-    flake-utils.lib.eachSystem [
-      "aarch64-linux"
-      "i686-linux"
-    ]
-      (system:
-        let pkgs = import nixpkgs {
-          inherit system;
-          overlays = self.overlays;
-        }; in
-        rec {
-          lib = {
-            nix-on-droid = { config }: import ./modules {
-              inherit pkgs config;
-              isFlake = true;
-              home-manager = (import home-manager { });
-            };
-          };
+  outputs = { self, nixpkgs, home-manager, flake-utils }: let
+    supportedSystems = [ "aarch64-linux" "i686-linux" ];
+  in flake-utils.lib.eachSystem supportedSystems (system: let
+    defaultPkgs = import nixpkgs {
+      inherit system;
+      overlays = [ self.overlay ];
+    };
+  in rec {
+    lib.nix-on-droid = { pkgs ? defaultPkgs, config }: import ./modules {
+      inherit pkgs home-manager config;
+      isFlake = true;
+    };
 
-          apps.nix-on-droid = flake-utils.lib.mkApp {
-            drv = (pkgs.callPackage ./nix-on-droid { });
-          };
-          defaultApp = apps.nix-on-droid;
-        }
-      );
+    apps.nix-on-droid = flake-utils.lib.mkApp {
+      drv = (defaultPkgs.callPackage ./nix-on-droid { });
+    };
+    defaultApp = apps.nix-on-droid;
+  }) // {
+    overlay = nixpkgs.lib.composeManyExtensions (import ./overlays);
+  };
 }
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -9,32 +9,30 @@
...
+  outputs = { self, nixpkgs, home-manager, flake-utils }: let
+    supportedSystems = [ "aarch64-linux" "i686-linux" ];
+  in flake-utils.lib.eachSystem supportedSystems (system: let
+    home-manager-input = home-manager;
+    defaultPkgs = import nixpkgs {
+      inherit system;
+      overlays = [ self.overlay ];
+    };
+  in rec {
+    lib.nix-on-droid = {
```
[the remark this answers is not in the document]
zhaofengli: Sure, let's do it in a separate PR.

flake.nix
```diff
@@ -9,32 +9,30 @@
...
+      pkgs ? defaultPkgs,
+      home-manager ? import home-manager-input { inherit pkgs; },
```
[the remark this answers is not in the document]
zhaofengli: Ah, I was misled by the impression that `home-manager` takes an initialized instance from the default value (`import <home-manager> {}`). Let's change it to accept a path-like instead so it's less confusing. Flake inputs are path-like so no special casing is required.

flake.nix
```diff
@@ -9,32 +9,24 @@
...
-          };
+  outputs = { self, nixpkgs, home-manager, flake-utils }: let
+    supportedSystems = [ "aarch64-linux" "i686-linux" ];
+  in flake-utils.lib.eachSystem supportedSystems (system: let
+    defaultPkgs = import nixpkgs {
+      inherit system;
+      overlays = [ self.overlay ];
+    };
+  in rec {
+    lib.nix-on-droid = { pkgs ? defaultPkgs, config }: import ./modules {
```
Gerschtli, line 20: Not necessary for this PR, but I would say, it is reasonable to add `home-manager` as argument.

modules/default.nix
```diff
@@ -16,7 +20,7 @@ let
   rawModule = evalModules {
     modules = [
       {
-        _module.args.home-manager = home-manager;
+        _module.args.home-manager = hmPath;
```
Gerschtli, line 23: If you change what is inside the `home-manager` argument, that could be a breaking change if anybody uses this argument in their configuration. But I don't really see a reason why anybody would do this.

Nonetheless, a name like `home-manager-path` would be better, because nobody would assume that `home-manager` is just a path.

flake.nix
```diff
@@ -9,32 +9,24 @@
...
+      inherit pkgs home-manager config;
```
Gerschtli, line 21: To only pass the path of home-manager, you could be more explicit.
```suggestion
      inherit pkgs config;
      home-manager = home-manager.outPath;
```

flake.nix
zhaofengli: Since `home-manager` now only provides the path and not any configuration, I don't see a need to add it here. Using native methods to override inputs in Flakes feels better to me.

Gerschtli: But you can not apply patches to the input natively with flakes but you could do it programmatically.

t184256:
Is this one a WIP or is it ready for review again?

zhaofengli:
> Is this one a WIP or is it ready for review again?

Just addressed the previous comments, and it should be ready for review again.

Gerschtli (approved):

t184256:
rebased slightly, only change is this omission in modules/home-manager.nix, rename inside a dropped block:

``` diff
--  tryEval = builtins.tryEval (home-manager.path + "/modules/modules.nix");
-+  tryEval = builtins.tryEval (home-manager-path + "/modules/modules.nix");
-   assertionNixpkgs = types ? submoduleWith;
-   assertionHomeManager = tryEval.success && builtins.functionArgs (import tryEval.value) ? useNixpkgsModule;
-   assertion = assertionNixpkgs && assertionHomeManager;
-
```

Thank you for your contribution.

PR #134, opened 2021-09-11
nix-on-droid.nix.default: fix the overlays inheritance of the home-manager config

ShamrockLee:
**Motivation**

Fix
```
error: attribute 'nixpkgs' missing
```

PR #147, opened 2021-12-05
Release 21.11

Gerschtli:
I prepared the release for 21.11, but for updating nixpkgs version there was an error because `pkgs/top-level/static.nix` was removed in https://github.com/NixOS/nixpkgs/commit/3edba5edfceef20c613c37967685f5eb04d1877d. I do not know that much about this cross-compiling stuff, will have a look at it but I am not sure if I am able to solve this.


t184256:
Awesome, I'll definitely have a look next week!

t184256:
Got a minute to look at it. https://github.com/NixOS/nix/commit/57409244ec8f568ce6b4ecb1bde0e509942ac05f#diff-112352307fc4c615843e2a683ed2dd10d73556e1c32932e0545270317d8fbe31 seems to turn pthreads initialization from warning to an error, breaking db initialization inside nix-directory.

But, this made me wonder, was it really necessary to use the target nix for it? Could we instead proot `-b`-mount `pkgsStatic.nix` someplace into the nix-directory building environment and use that to init? If yes, we're trading the hackily pulled in qemu-user binary for a `pkgsStatic.nix`.

I'll try that today or tomorrow if you don't get to it first.

t184256:
> there was an error because pkgs/top-level/static.nix was removed in NixOS/nixpkgs@3edba5e

I believe it was an overlay of hacks to make stuff link statically, not having it is not alarming.

> I do not know that much about this cross-compiling stuff, will have a look at it but I am not sure if I am able to solve this.

It sure looks like you've succeeded =)

t184256:
OK, today I'm stopping at `experimental Nix feature 'nix-command' is disabled` during bootstrap; most likely `nix build` in login inner meets one of the Nix 2.4 breakages.

Gerschtli:
Oh right, because also the nix-on-droid script uses the new nix command, this feature should be enabled per default.

Gerschtli:
I think the main question is: Do we want to use nix 2.4 as default or stick with nix 2.3 as nixpkgs-stable defaults to. Maybe it would be easier to keep it at nix 2.3 for now and document (or implement it by default) that `experimental-features = nix-command` needs to be enabled.

Gerschtli:
Yes, so I removed the nix update but added support if someone wants to enable nix 2.4 so that `nix-on-droid` script still works.

t184256:
Installs fine now. My next steps are, I guess, a rebase of the tests PR on top of this, testing and careful reading. Got any more plans for the release?

Gerschtli:
No, that would be all for this PR.

Gerschtli:
One thing came to my mind: Maybe this would be a good opportunity to get rid of these assertions in the home-manager module, asserting a newer version of nixpkgs and home-manager than 19.09.

modules/environment/nix.nix
```diff
@@ -75,6 +91,7 @@ in
         "cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY="
         "nix-on-droid.cachix.org-1:56snoMJTXmDRC1Ei24CmKoUqvHJ9XCp+nidK7qkMQrU="
       ];
+      experimentalFeatures = mkIf (versionAtLeast cfg.package.version "2.4") [ "nix-command" ];
```
t184256, line 94: https://nixos.org/manual/nix/stable/command-ref/conf-file.html says it's enabled by default in stable (2.4), do we still need to worry about it?

t184256: or is this exactly why we have to worry about it? =)

Gerschtli: Fun fact: Yesterday the default was 0, I wondered what that meant.. Seemed to be a bug in the documentation :D Then we definitely do not need this, I will remove it.

modules/environment/login/nix-on-droid.nix.default
```diff
@@ -28,23 +28,25 @@
...
   #home-manager.config =
   #  { pkgs, lib, ... }:
   #  {
   #    # Read the changelog before changing this value
-  #    home.stateVersion = "21.05";
+  #    home.stateVersion = "21.11";
   #
   #    # Use the same overlays as the system packages
-  #    nixpkgs.overlays = lib.mkIf (lib.hasAttrByPath [ "nixpkgs" "overlays" ] config) config.nixpkgs.overlays;
+  #    nixpkgs = { inherit (config.nixpkgs) overlays; };
```
t184256, line 44: Is the #134 case covered enough by the on-device flake test?

Gerschtli: Now I finally understood the whole issue and why this fix was needed. I need to do some changes in `modules/nixpkgs.nix` to let this one work. (When in a flake setup, you should not be allowed to change the `pkgs` instance, but the option should be available to be at least able to evaluate the config.)

To your question: No, we should have a test case for home-manager integration for flake and non-flake setup. To reproduce the issue of #134, you should set `nixpkgs.overlays = [ ];` in a flake setup, I think.

t184256: Not sure if 8511636 has resolved it, getting the same #134 symptoms...

t184256 (approved):
I believe it's fine. Thanks for your work, much appreciated!

Gerschtli:
Please do not forget to push the new bootstrap to https://nix-on-droid.unboiled.info/ :)

Gerschtli:
In case you didn't see my comment, ping @t184256 :)

t184256:
I do remember, I haven't actually tagged 21.11 yet. I planned to also pick up #150, #153 maybe, some assorted fixes to README and dropping i686. And if you plan any more exciting stuff, please warn me, I like our current practice of big releases and backporting basically never.

Gerschtli:
Just to be clear, we never tag and only create branches for the release from that commit where we drop support for the old nixpkgs and home-manager releases, at least that was my understanding. So I think branching `release-21.11` off starting with b30833b21045397b3c8c10414181d47839810607 (removal of make workaround -> breaking change) would be a good idea.

Wait, I just noticed that our branches are a bit confusing :D So currently after releasing the version, the release branch does not get any updates? This forces every user to follow master if they want updates more often than the nixpkgs releases. Didn't realise that until now. It works for the current (I would estimate) small user base of nix-on-droid, but maybe it needs some refinement in the future.

So either way, let's stay with the release idea.

I like the idea to drop i686, reduces development overhead. I think we could give the changelog a bit more love, but other than that I think the release looks complete.

How do you think about keeping track and gathering bigger things we want to accomplish? I saw that github has some kind of a ticket board (trello like), or maybe we use the discussions for that. Because sometimes I have so many ideas and want to write them down somewhere, I think that would help sync ourselves to common goals or to plan things better.

BTW: I love working on this project with you, always makes a lot of fun to see what we creating here :)

t184256:
> Just to be clear, we never tag and only create branches for the release

True; wrong term; my bad. I meant that I like our branches mostly-immovable (and mostly-fast-forwarding-when-we-do-move-them).

> from that commit where we drop support for the old nixpkgs and home-manager releases, at least that was my understanding

I didn't know we had a system =) Mine was way less defined: "release preparation activity dies down, I create the branch and update app".

> the release branch does not get any updates?

I think we sometimes merge --ff-only really innocent things, but when the first major thing lands, the branch stops advancing and we [tell people to use master](https://github.com/t184256/nix-on-droid/pull/132#issuecomment-922496859).

> This forces every user to follow master if they want updates more often than the nixpkgs releases. Didn't realise that until now. It works for the current (I would estimate) small user base of nix-on-droid, but maybe it needs some refinement in the future.

I'd say it works for our model of dropping big changes around nixpkgs release time.

> How do you think about keeping track and gathering bigger things we want to accomplish? I saw that github has some kind of a ticket board (trello like), or maybe we use the discussions for that. Because sometimes I have so many ideas and want to write them down somewhere, I think that would help sync ourselves to common goals or to plan things better.

Sounds good, would be more helpful than syncing plans in MR comments. I think we had a discussion for such planning once, and this makes even more sense than that.

> BTW: I love working on this project with you, always makes a lot of fun to see what we creating here :)

Thanks, the feeling is mutual, and I dare say it won't move even at a quarter of its current pace if it were not for you. I have a secret plan to up my Nix game this winter, so hope I'll be of more use after that.

Gerschtli:
I agree, the way we handle releases is fine and communicated correctly.

> I think we had a discussion for such planning once, and this makes even more sense than that.

So you are in favor of this board or of the discussions feature of github? I would be fine with both, I never worked with either of them.

> Thanks, the feeling is mutual, and I dare say it won't move even at a quarter of its current pace if it were not for you. I have a secret plan to up my Nix game this winter, so hope I'll be of more use after that.

Thank you, sometimes it feels like I create multiple big PRs in a week and then a couple of months nothing, but you are always fast in regards of reviews and responses. Whatever "to up my Nix game" is but it sounds promising. Flakifying my whole setup was the motivation I needed to get my hands dirty again on this project :)

t184256:
Tried some fancy board thing, let's see how useful it is.

PR #149, opened 2021-12-08
pkgs/nix-directory: stop using qemu-user to init nix db

t184256:
@Gerschtli  Okayy, so, turns out I've been using qemu all that time to init Nix DB with target Nix instead of the host Nix for... no good reason whatsoever? Dropped qemu, initialized it with a static Nix, checked that it works alright with x86_64 as the host and aarch64 as the target, got what I think is a good stepping stone towards resolving the nix update in https://github.com/t184256/nix-on-droid/pull/147


pkgs/nix-directory.nix
```diff
@@ -56,9 +52,12 @@ stdenv.mkDerivation {
       cp --archive "$i" "${buildRootDirectory}$i"
     done
 
-    USER=${config.user.userName} ${prootCommand} "$PKG_NIX/bin/nix-store" --init
-    USER=${config.user.userName} ${prootCommand} "$PKG_NIX/bin/nix-store" --load-db < .reginfo
-    USER=${config.user.userName} ${prootCommand} "$PKG_NIX/bin/nix-store" --load-db < ${prootTermuxClosure}/registration
+    ls -l /proc/self/maps
+    cat /proc/self/maps > /dev/null
```
Gerschtli, line 56: What is `/proc/self/maps` and why are ls'ing and piping it to `/dev/null`? I think a short comment would for future reference here :)

pkgs/default.nix
```diff
@@ -44,7 +44,6 @@ let
     bootstrap = callPackage ./bootstrap.nix { };
     bootstrapZip = callPackage ./bootstrap-zip.nix { };
     prootTermux = callPackage ./cross-compiling/proot-termux.nix { };
-    qemuAarch64Static = callPackage ./qemu-aarch64-static.nix { };
```
Gerschtli, line 47: Shouldn't the copyright header be adjusted?

BTW: This is always a bit tedious, not sure but maybe https://github.com/fsfe/reuse-tool could help dealing with this stuff in a less distracting way. Didn't you it yet but heard good things about it :)

Gerschtli (approved):
That is pretty slick! Seems pretty obvious in hindsight :D Good catch :)

pkgs/nix-directory.nix
t184256: oh wait, that's some debugging stuff I need to drop. it was there just to confirm /proc/self is there and

pkgs/default.nix
t184256: Adjusted.

I think I've evaluated that tool when playing with https://github.com/t184256/python-nix-project-template, but I don't remember why I ended up not using it there. Yeah, the manual updating when we remember to workflow is not ideal, solutions are welcome.

PR #151, opened 2021-12-12
tests/fakedroid: emulate a nix-on-droid-like env for testing

t184256:
Run "on-device-tests" in CI. It sure won't catch some of the weirder bugs, but it'll catch the simpler regressions.

Had to dump everything it builds to cachix, lest it takes an hour+ to run. Which is a bit weird given that on my laptop it runs comparable to my phone, I guess github actions is slower than both.

Funny how we take Ubuntu with Nix and emulate what the android app does to then run Nix inside it to test a project that brings Linux/Nix experience to... a fake not-android Linux/Nix setup.

@Gerschtli, I've tried to annotate the script, hope it's more readable now, suggestions are welcome.


.github/workflows/fakedroid-odt.yml
```diff
@@ -0,0 +1,38 @@
...
+      with:
+        name: nix-on-droid
+        signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
+
+    - name: init
+      run: tests/fakedroid.sh echo INIT
+
+    - name: init-upload
+      if: always()
+      run: tests/fakedroid.sh nix-shell -p cachix --run 'nix path-info --all | cachix push nix-on-droid'
```
Gerschtli, line 28: Why do we need two push steps in the pipeline?

tests/proot-test.nix
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) 2019-2021, see AUTHORS. Licensed under MIT License, see LICENSE.
+
+{ pkgs ? import <nixpkgs> { } }:
+
+let
+  #stdenv = pkgs.pkgsStatic.stdenvAdapters.makeStatic pkgs.stdenv;
+  stdenv = pkgs.stdenv;
+in
+
+stdenv.mkDerivation {
```
Gerschtli, line 10: This proot should be the same version + nixpkgs rev like the on-device one except for the static and cross compiling stuff. Maybe for now you can just somehow import the definition from `pkgs/` to keep it consistent.

tests/fakedroid.sh
```diff
@@ -0,0 +1,114 @@
...
+    # Build a zipball:
+    nix build --show-trace -f pkgs \
+        --argstr arch aarch64 \
+        --argstr nixOnDroidChannelURL file:///n-o-d/archive.tar.gz \
+        bootstrapZip -o .fakedroid/inj/nix-on-droid-aarch64
+    ZIPBALL=$(realpath .fakedroid/inj/nix-on-droid-aarch64/bootstrap-aarch64.zip)
+    # Unpack the zipball the way the Android app does it:
+    pushd .fakedroid/env/$INSTALLATION_DIR
+        unzip "$ZIPBALL"
+        chmod -R 700 .
```
Gerschtli, line 71: Why should every file be executable?

tests/fakedroid.sh
```diff
@@ -0,0 +1,114 @@
...
+            SYM_TGT=${e%%←*}
+            SYM_SRC=${e##*←}
+            ln -sf "$SYM_TGT" "$SYM_SRC"
+        done < SYMLINKS.txt
+        while read e; do
+            chmod +x "$e"
+        done < EXECUTABLES.txt
+        rm SYMLINKS.txt EXECUTABLES.txt
+    popd
+    chmod -R 700 .fakedroid/env
```
Gerschtli, line 82: Same question

tests/fakedroid.sh
```diff
@@ -0,0 +1,114 @@
...
+# Inject nix-on-droid version under test into the environment.
+# Uncommitted chages won't be picked up, just HEAD.
+# /n-o-d/archive.tar.gz is used as a channel, /n-o-d/unpacked --- as a flake.
+
+rm -rf .fakedroid/env/n-o-d; mkdir -p .fakedroid/env/n-o-d/unpacked
+git archive --format=tar --prefix n-o-d/ HEAD \
+    > .fakedroid/env/n-o-d/archive.tar
+tar --strip-components=1 -xf \
+    .fakedroid/env/n-o-d/archive.tar -C .fakedroid/env/n-o-d/unpacked
+gzip .fakedroid/env/n-o-d/archive.tar
```
Gerschtli, line 95: Not sure if the gzip is necessary. Nix should be able to handle .tar files. But didn't test it.

tests/fakedroid.sh
```diff
@@ -0,0 +1,114 @@
...
+
+
+# The "first boot" proot invocation:
+
+SH=$(readlink .fakedroid/env/$INSTALLATION_DIR/bin/sh)
+
+env -i $PASSTHROUGH_VARS $PROOT $PROOT_ARGS \
+    $INSTALLATION_DIR/$SH ${INSTALLATION_DIR}/usr/lib/login-inner <<<'echo OK'
```
Gerschtli, line 103: Why these `<<<`?

tests/on-device/lib.bash
```diff
@@ -11,11 +11,14 @@ setup() {
...
-    REPO_NAME=${BASH_REMATCH[2]}
-    REPO_BRANCH=${BASH_REMATCH[3]}
-    FLAKE_URL=github:$REPO_USER/$REPO_NAME/$REPO_BRANCH
+    if [[ "$CHANNEL_URL" =~ https://github.com/(.+)/(.+)/archive/(.+)\.tar\.gz ]]; then
+      REPO_USER=${BASH_REMATCH[1]}
+      REPO_NAME=${BASH_REMATCH[2]}
+      REPO_BRANCH=${BASH_REMATCH[3]}
+      FLAKE_URL=github:$REPO_USER/$REPO_NAME/$REPO_BRANCH
+    elif [[ "$CHANNEL_URL" == file:///n-o-d/archive.tar.gz ]]; then
+      FLAKE_URL=/n-o-d/unpacked
```
Gerschtli, line 20: Does this directory work as a flake without a git repo in it?

tests/fakedroid.sh
t184256: It is, because nix-channel's heuristics require `.tar.*` to recognize it's a tarball. =/

Gerschtli:
I think the script is good, I just see a lot of duplication of some parts (e.g. proot mounts, symlinks/executables, login-inner mv). Hopefully at some point, we can do some nixification of this script to reduce this duplication, but for now totally fine. As long as we have a working setup to test on our computers (which is really incredible!).

tests/fakedroid.sh
t184256: Because the initial script execs /bin/bash unconditionally, that's a way to keep it non-interactive.

tests/on-device/lib.bash
t184256: Yes, flake doesn't have to be an initialized repo, it's just the behaviour changes significantly when it is.

.github/workflows/fakedroid-odt.yml
t184256: We don't, that's from the time I didn't know how to `if: always()`. Removed one.

tests/fakedroid.sh
t184256: That's not right, I think I was aiming for u+rw here.

Gerschtli: Why would that be needed in the first place?

tests/fakedroid.sh
Gerschtli: Damn it..

tests/on-device/lib.bash
Gerschtli: That is good to know! :)

Gerschtli:
From the reviewers POV it would help a lot if you would push separate commits instead of overwriting the same commit again. After the PR review is finished you can rebase your commits. That way I would actually be able to see, what you changed :)

tests/fakedroid.sh
t184256: Because unzip unpacks with -r-xr-xr-x: https://github.com/t184256/nix-on-droid/runs/4497615108?check_suite_focus=true#step:5:6223

t184256:
OK, I'll try to remember this isn't GitLab where it's not a problem.

tests/fakedroid.sh
Gerschtli: Ah right makes sense, it was build from stuff coming from /nix/store.

tests/proot-test.nix
t184256: Indeed. Extracted common stuff into `pkgs/proot-termux` used from both places, hope I've also passed the same nixpkgs as well.

tests/fakedroid.sh
t184256: Added some comment, removed second chmod.

modules/environment/login/default.nix
```diff
@@ -85,8 +85,8 @@ in
       prootStatic =
         let
           crossCompiledPaths = {
-           aarch64 = "/nix/store/45jqhgc0cg1l6vrl95p3qg3j4wh6w59y-proot-termux-aarch64-unknown-linux-android-unstable-2021-11-21";
-           i686 = "/nix/store/8nww78ndkbxqhi4h9f0ly9zkwixa0cd9-proot-termux-i686-unknown-linux-android-unstable-2021-11-21";
+           aarch64 = "/nix/store/dapbgzbpl426jrhz4a2sdl096a8l98ad-proot-termux-aarch64-unknown-linux-android-unstable-2021-11-21";
+           i686 = "/nix/store/6nnjhrh8pgyxjnya72irahxpkbnxai1w-proot-termux-i686-unknown-linux-android-unstable-2021-11-21";
```
Gerschtli, line 89: Out of curiosity: Why did these change?

tests/proot-test.nix
```diff
@@ -3,33 +3,9 @@
 { pkgs ? import <nixpkgs> { } }:
```
Gerschtli, line 3: That is still not the same nixpkgs. Is a bit of nitpicking, so I am fine with it, at least it is the same version of termux :)

modules/environment/login/default.nix
t184256: I don't know, actually. The -O3 I've added isn't to blame.

tests/proot-test.nix
t184256: Yeah, I pushed that fix to the wrong branch, should be `(import ../pkgs/lib/load-nixpkgs.nix {})`.

modules/environment/login/default.nix
Gerschtli: It isn't? It is a change in the inputs of the derivation and (unless being the default) changes the output.

(Didn't see the `-O3` there before)

t184256: Yeah,  but I couldn't figure out what have I changed...

Gerschtli: I'm not sure I understand what you mean.. If you add `CFLAGS`, this is what you changed. :D

t184256: I meant, it changed before I've added CFLAGS, and the best explanation I have is the renaming of the patch.

Gerschtli: Ah gotcha. Yeah that also sounds reasonable, because you are using the path not the content of the patch file as derivation input.

tests/fakedroid.sh
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env bash
+
+# This is a script to run "on-device" tests in CI, without the device.
+# Takes the bootstrap aarch64 zipball, unpacks, proots into it.
+# This won't catch all bugs, of course, but that's something.
+
+# Set up envvars, prepare directories:
+
+set -uexo pipefail
```
Gerschtli, line 9: Could you remove the -x or at least make it optional? I tested it locally and it is producing a lot of noise when you are not interested in changing the fakedroid.sh script.

Gerschtli:
Tested locally, runs perfectly fine. Please add `/.fakedroid` to .gitignore :)

tests/fakedroid.sh
t184256: Sure, added

Gerschtli (approved):

PR #153, opened 2021-12-14
Fix problems with shell init when logging in via ssh

Gerschtli:
Load session-init in shell init files (`/etc/profile` for bash, `/etc/zshenv` for zsh).

Remove `-u` in session-init because it would enable `-u` when it is loaded as part of a shell init config file.

Supersedes #140.

Ping @azuwis


modules/environment/shell.nix
```diff
@@ -0,0 +1,32 @@
...
+  };
+
+
+  ###### implementation
+
+  config = {
+
+    environment.etc = {
+      "profile".text = ''
+        . "${config.build.sessionInit}"
```
t184256, line 22: It's a directory containing `etc/profile.d/nix-on-droid-session-init.sh`.

modules/environment/shell.nix
```diff
@@ -0,0 +1,32 @@
...
+      '';
+
+      "zshenv".text = ''
```
t184256, line 25: Can we not go the 'every shell' way, but instead put it into some `~/.ssh/rc`?

t184256 (changes requested):
+ General question: is there value in bypassing that initialization for ssh login (e.g., condition it on an AcceptEnv'ed variable)?

modules/environment/shell.nix
azuwis: NixOS handles every shell, because every shell is different.

Bash https://github.com/NixOS/nixpkgs/blob/ef3a8a853b82db0908870a62ad91d03c36ca4331/nixos/modules/programs/bash/bash.nix#L121

Fish https://github.com/NixOS/nixpkgs/blob/3d718c64b595e914e7539509846457fe898ad31a/nixos/modules/programs/fish.nix#L176

Zsh https://github.com/NixOS/nixpkgs/blob/ef3a8a853b82db0908870a62ad91d03c36ca4331/nixos/modules/programs/zsh/zsh.nix#L177

But they all check the same env var `$__NIXOS_SET_ENVIRONMENT_DONE`.

azuwis: Also, `~/.ssh/rc` is specific to openssh, it will not work for dropbear, mosh, telnetd, etc.

I think handling session init in shell rc files is more consistent, NixOS and nix-darwin both go this way.

modules/environment/shell.nix
Gerschtli: Oh right. Then why did it work for me.. I will have a look at that..

modules/environment/shell.nix
Gerschtli: I would agree to follow upstream's approach. But I don't have much knowledge about ssh initialisation process. Either way adding that to the shell init script does not hurt because this file is only loaded once.

Gerschtli:
TBH I can not think of a reason to not load session-init on ssh login.

modules/environment/shell.nix
t184256: OK, guess this *is* the easy way.

modules/environment/shell.nix
Gerschtli: Fixed the description and type of sessionInit option :)

t184256: Thank you.

t184256 (approved):

PR #154, opened 2021-12-14
home-manager: add extraSpecialArgs and sharedModules options

Gerschtli:
To reach feature parity with the NixOS and nix-darwin modules.

### DIFF
--- a/modules/home-manager.nix
+++ b/modules/home-manager.nix
@@ -10,7 +10,7 @@ let
   extendedLib = import (home-manager-path + "/modules/lib/stdlib-extended.nix") pkgs.lib;
 
   hmModule = types.submoduleWith {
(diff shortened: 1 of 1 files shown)


t184256 (approved):
r+

I don't find `specialArgs`' example descriptive, but I guess it's a question to upstream.
`sharedModules`' name's a bit weird in our context, but let's keep it.

Let's also agree that you can merge reviewed PRs as you see fit, OK?

Gerschtli:
Yeah I agree to your comments, but wanted to be in sync with upstream so that in the future that whole home-manager module could be easily moved to the home-manager repo.

> Let's also agree that you can merge reviewed PRs as you see fit, OK?

I am fine with that :)

PR #157, opened 2021-12-14
ci: do not attempt to push to cachix on PR builds

Gerschtli:
Secrets are not available when triggered by a PR. Therefore, `CACHIX_SIGNING_KEY` is not defined and cachix will always fail.

### DIFF
--- a/.github/workflows/fakedroid-odt.yml
+++ b/.github/workflows/fakedroid-odt.yml
@@ -30,5 +30,5 @@ jobs:
         tests/fakedroid.sh nix-on-droid on-device-test
 
     - name: upload
-      if: always()
+      if: always() && github.event_name != 'pull_request'
       run: tests/fakedroid.sh nix-shell -p cachix --run 'nix path-info --all | cachix push nix-on-droid'


.github/workflows/fakedroid-odt.yml
```diff
@@ -30,5 +30,5 @@ jobs:
         tests/fakedroid.sh nix-on-droid on-device-test
 
     - name: upload
-      if: always()
+      if: always() && github.event_name != 'pull_request'
```
t184256, line 33: Hope it can be shortened to just `github.event_name != 'pull_request'`.

Gerschtli: Good question, I am not sure, I will test it with this PR.

Gerschtli: `always()` does more magic than just returning `true`: https://github.com/t184256/nix-on-droid/runs/4526292325?check_suite_focus=true

But `continue-on-error` would work: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

t184256: Yeah, I take my suggestion back O.o

Issue #162, opened 2021-12-17
Installation upgraded to 21.11 complains about the lack of /dev/shm


t184256:
Installations upgraded from 21.05 to 21.11 give the following warning on the next "boot":
`proot warning: can't sanitize binding "/data/data/com.termux.nix/files/usr/dev/shm": No such file or directory`. Newly-bootstrapped setups are not affected.

The manual workaround is to `mkdir -p /data/data/com.termux.nix/files/usr/dev/shm` and restart the app.

PR #163, opened 2021-12-17
modules/build/activation.nix: fix missing /dev/shm on upgrade

t184256:
Fixes: #162

I plan for this to be the last MR going into the release, I haven't find other issues during upgrade testing.

Activation-time quick fix was chosen over login-time quick fix because I can imagine a hot loop around login, but activation is slow in comparison.

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -17,6 +17,7 @@
   ./home-manager.nix
   ./nixpkgs/options.nix
   ./time.nix
+  ./upgrade.nix
   ./user.nix
   ./version.nix
   (pkgs.path + "/nixos/modules/misc/assertions.nix")
--- a/modules/upgrade.nix
+++ b/modules/upgrade.nix
@@ -1,0 +1,15 @@
+# Copyright (c) 2019-2021, see AUTHORS. Licensed under MIT License, see LICENSE.
+
+{ config, lib, ... }:
+
+{
+  config.build.activationAfter =
+    # TODO: remove when we stop supporting upgrades from <21.11
+    # Setups upgraded to 21.11 don't have the /dev/shm directory bootstrapped:
+    # https://github.com/t184256/nix-on-droid/issues/162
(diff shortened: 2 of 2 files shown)


Gerschtli:
I do not really get the source of this problem. How can it be that this directory is only missing sometimes and why can we remove this after 21.11?

I think a test for stateVersion < 21.11 would be better then and only run the fix for these cases?

t184256:
> I do not really get the source of this problem. How can it be that this directory is only missing sometimes

The only place it's created in currently is inside the zipball https://github.com/t184256/nix-on-droid/blob/149e5c886999cacfecdab639878865f833408742/pkgs/bootstrap.nix#L10

If you install clean, you get it from zipball unpacking. If you bootstrap on <21.11 and upgrade, it's never created, thus the warning.

> and why can we remove this after 21.11?

Because anyone ever activating 21.11 will have it.

> I think a test for stateVersion < 21.11 would be better then and only run the fix for these cases?

That didn't occur to me, could you propose a fix? Users are supposed to first activate with old stateVersion and only then bump it, right?

Gerschtli:
Got it.

I think a simple mkIf should do it like:
```nix
activationAfter = mkIf (system.stateVersion <some-less-than-op> "21.11") { mkdirDevShm = "..."; };
```
For the `<some-less-than-op>` you could grep through nixpkgs for the correct name.

t184256:
Ah. OK, I figured it out. Tested to upgrade w/o warning and to remove createDevShm after bumping stateVersion.

Gerschtli (approved):

t184256:
And here's where I place release-21.11 and push the app update.

Go grab a beverage of your choice and celebrate, @Gerschtli, big thanks for all the [code](https://github.com/t184256/nix-on-droid/graphs/contributors) and reviews you've put into it.

Gerschtli:
Nice, thank you too! Like always a lot of fun. Hope some time we can grab one together :)

PR #165, opened 2021-12-20
Create fake /proc/stat to make unpached htop work

azuwis:


t184256:
That's an interesting approach. Can you come up with cases where a failure would be preferred to a fake file?

What I can criticize about the implementation is that it can mask a working `/proc/stat` if login is running under root. Could you condition masking it on unreadability of `/proc/stat`? Also maybe suffix the fake filename with `-fake` for extra clarity.

azuwis:
Please review the new changes.

modules/upgrade.nix
```diff
@@ -11,5 +11,11 @@
       createDevShm = ''
         mkdir -p ${config.build.installationDir}/dev/shm
       '';
+      createProcStat = ''
```
t184256, line 14: There already can be installs on 21.11 that don't have the file created, so the condition isn't fully accurate. But maybe we don't need to iron this out, see below.

modules/upgrade.nix
```diff
@@ -11,5 +11,11 @@
...
+        mkdir -p "${config.build.installationDir}/proc"
+        if [ ! -e "${config.build.installationDir}/proc/stat-fake" ]; then
+          echo "btime 0" > "${config.build.installationDir}/proc/stat-fake"
```
t184256, line 17: I'm wary of managing content that way, working around Nix. What if we'll want to later extend it with more fields for some other software?

What about taking a different approach: make it a config file in `/etc` by adding one more module to `modules/environment`? That way there will be no need to create `/proc` and no upgrade concerns, plus configurability and overrideability through Nix. And in login we can check for "unreadable real file *and* present fake file => bind-mount it as /proc/stat". Does that sound reasonable?

One quirk I foresee though is that this `/etc/proc-stat-fake` will be a symlink that doesn't make sense outside of proot, so getting to the file we want to bind-mount will require prefixing, something like `"$installationDir/$(readlink $installationDir/etc/proc-stat-fake)"`.

t184256:
Hm, indeed, that should be enough if you don't need it non-root only =)

azuwis:
Please review the new changes.

t184256 (approved):
Great, I really like this version. Terse, upgradeable, seemingly all-around safe and probably also better than all my suggestions, actually. Thanks for tirelessly reworking it.

t184256:
Yeah, good thinking. It was also frozen at an old version.

t184256:
Removed in c1b61fc (should be trivial enough to not warrant a review).

Issue #166, opened 2021-12-30
NixOS module compatibility


jrobsonchase:
Complete compatibility obviously isn't possible, but for things like auto-generated cachix configurations, it would be convenient to reuse them in nix-on-droid.

Continuing with the cachix config example, NixOS has `nix.binaryCaches` and `nix.binaryCachePublicKeys`, while nix-on-droid has the equivalent `nix.substituters` and `nix.trustedPublicKeys`. Instances such as these where attrs have the same type and semantics should favor the NixOS names rather than create new incompatible ones that break compatibility.

Gerschtli:
We do not break compatibility because we never wanted that between NixOS and nix-on-droid. WIth your examples: We aligned to the new wording of latest nix releases, NixOS just still uses the deprecated namings. So we will not rename option to a deprecated wording. Ideally these options should be renamed in NixOS.

Like you said, there will never be compatibility between NixOS and nix-on-droid but if you would like to see new options added to nix-on-droid, feel free to open an issue or a PR.

jrobsonchase:
> We aligned to the new wording of latest nix releases, NixOS just still uses the deprecated namings. So we will not rename option to a deprecated wording. Ideally these options should be renamed in NixOS.

While I agree in principle, from a practical perspective, getting that changed in NixOS is much more difficult than adding compatibility here. Big, slow-moving projects, etc. etc.

I've used the `mkRenamedOption` pattern in home-manager modules to add new names for options while keeping deprecated support for the old name around. Would it not be possible to retcon the NixOS options as the "old" name in cases like these?

Gerschtli:
I do not get why this would be that difficult in NixOS? Just plug in a `mkRenamedOption`. I can create the PR (if not already one present) in a few days.

In the mean time you can add the `mkRenamedOption` in your own config if you need these aliases but I do not see the big benefit because you will never be able to share your whole config between NixOS and nix-on-droid and if we start with these two options then we should be somehow consistent with this compatibility approach which does not look reasonable to me.

jrobsonchase:
And will such a change in NixOS make it back to the stable release channels as well? I'm not super familiar with where option renames are allowed to land stability-wise. If it's just the `-unstable` branches and the next stable release, then the upstream rename is a non-starter as far as this issue is concerned. It means that anyone wanting to share configs that *should* be compatible won't be unable to until the next release (and they actually migrate to it), unless they're using `unstable`. Even then, it still forces migration from the user.

[...]

PR #167, opened 2021-12-30
module/nix: Add backward compat with nixos for binary caches

jrobsonchase:
Per #166, complete compatibility with NixOS is a non-goal, and nix-on-droid options won't be renamed to match.

But it won't hurt to add compatibility where possible, and it's much simpler to do so here than to try to make the rest of the world change.

### DIFF
--- a/modules/environment/nix.nix
+++ b/modules/environment/nix.nix
@@ -6,9 +6,16 @@ with lib;
 
 let
   cfg = config.nix;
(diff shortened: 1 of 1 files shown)


Gerschtli:
If @t184256 does not have any objections, we can merge that.

t184256 (approved):
Looks tame and doesn't seem hard to maintain.

PR #178, opened 2022-04-11
Add options `nix.nixPath` `nix.registry`

azuwis:
Also rename `nix.extraConfig` to `nix.extraOptions`

Most of the codes are copied from NixOS modules.

### DIFF
--- a/modules/environment/nix.nix
+++ b/modules/environment/nix.nix
@@ -46,6 +46,65 @@ in
         '';
       };
 
+      registry = mkOption {
+        type = types.attrsOf (types.submodule (
+          let
+            referenceAttrs = with types; attrsOf (oneOf [
+              str
+              int
+              bool
+              package
+            ]);
+          in
+          { config, name, ... }:
+          {
+            options = {
+              from = mkOption {
+                type = referenceAttrs;
+                example = { type = "indirect"; id = "nixpkgs"; };
+                description = "The flake reference to be rewritten.";
+              };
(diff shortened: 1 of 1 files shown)


modules/environment/nix.nix
```diff
@@ -63,28 +73,32 @@ in
...
-      '';
-    };
+  config = mkMerge [
+    {
+      environment.etc = {
+        "nix/nix.conf".text = ''
+          sandbox = false
+          substituters = ${concatStringsSep " " cfg.substituters}
+          trusted-public-keys = ${concatStringsSep " " cfg.trustedPublicKeys}
+          ${cfg.extraConfig}
```
t184256, line 142: needs to be renamed, or there will be warnings by default

t184256:
Looks OK to me.

1. one renaming needed, comment inline
2. please bump the year
3. ideally mimic style of other commit messages
4. it looks largely lifted from nixpkgs, please add attribution. smth like

```
# Based on
# https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/misc/nix-daemon.nix
# (Copyright (c) 2003-2022 Eelco Dolstra and the Nixpkgs/NixOS contributors,
#  licensed under MIT License as well)
```

t184256 (changes requested):

azuwis:
Please review the new changes.

t184256 (approved):
Thank you!

PR #179, opened 2022-04-20
modules/environment/login: Fix fakeProcStat path

azuwis:


t184256:
And here I thought it's trivial enough for me to not break anything...
Thanks for catching this regression!

t184256 (approved):

PR #189, opened 2022-07-09
prerelease-22.05-pt2

t184256:
Use non-prebuilt SDK and attempt to unhardcode a path to proot.

@Gerschtli, if you got time, could you please help me with straightening it out? I know what I want to do here, but not how to do it right. =(

### DIFF
--- a/modules/environment/login/default.nix
+++ b/modules/environment/login/default.nix
@@ -82,7 +82,7 @@ in
     environment.files = {
       inherit login loginInner;
 
-      prootStatic = "/nix/store/scj2cqyvcra7dcan4y73n0459ac7fms7-proot-termux-aarch64-unknown-linux-android-unstable-2022-05-03";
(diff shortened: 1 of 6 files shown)


Gerschtli:
Not sure if I improved it, I don't know how to make this all less strange TBH. It is just that we have 3 different ways how to build proot (bootstrap, on-device, fakedroid), that's why I tried to move this complexity to one file `pkgs/proot-termux/default.nix`.

t184256:
Maybe axing the bootstrap one and going through proot would cut down the complexity? Will have to fix the proot one first though.

t184256:
OK, I guess it's time to admit that I, again, don't have the energy to solve these puzzles in reasonable time. Trimmed the switching to non-precompiled proot, so that we can release at least the SDK change.

Gerschtli (approved):
Sounds good to me! :)

t184256:
Pushed as release-22.05.

PR #196, opened 2022-09-03
Update actions, apply coding style

Gerschtli:
Found new command `nix fmt` and https://github.com/nerdypepper/statix, quite nice to catch some inconsistencies and stuff. Also updated the github workflows. What do you think?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -45,5 +45,8 @@
         default = app;
         nix-on-droid = app;
       };
-    };
+    }
+    // flake-utils.lib.eachDefaultSystem (system: {
+      formatter = nixpkgs.legacyPackages.${system}.nixpkgs-fmt;
+    });
 }
--- a/flake.nix
+++ b/flake.nix
@@ -2,14 +2,15 @@
   description = "Nix-enabled environment for your Android device";
 
   inputs = {
-    flake-utils.url = "github:numtide/flake-utils";
+    nixpkgs.url = "github:NixOS/nixpkgs";
     home-manager = {
       url = "github:nix-community/home-manager";
       inputs.nixpkgs.follows = "nixpkgs";
     };
+    flake-utils.url = "github:numtide/flake-utils";
   };
 
-  outputs = { self, nixpkgs, flake-utils, home-manager }:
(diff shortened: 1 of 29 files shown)


t184256 (commented):
Formatting: opinionated, but OK

Actions: r+

flake.lock: why is nixpkgs pinned to a path?

Gerschtli:
> Actions: r+

what does r+ mean?

> flake.lock: why is nixpkgs pinned to a path?

my bad, didn't want to commit that one

t184256:
> what does r+ mean?

positive review

Gerschtli:
Pipeline is having trouble with `flake.lock`. Some general thought to that: How about adding a `flake.lock` because when we start building the bootstrap zip with flake setup we will either way add a lock file and when using nix-on-droid as input, one should always override nixpkgs and home-manager inputs, so should not break anything and would improve portability because nix tools always expect a flake.lock. What do you think?

t184256:
> one should always override nixpkgs and home-manager inputs

I dislike the consequences of forgetting to do that, but I do agree that avoiding adding a flake.lock is not a proper solution.

I guess we can add it.

Gerschtli:
@t184256 PTAL

flake.nix
```diff
@@ -40,7 +41,13 @@
...
-      defaultApp.aarch64-linux = app;
-    };
+
+      apps.aarch64-linux = {
+        default = app;
+        nix-on-droid = app;
+      };
+    }
+    // flake-utils.lib.eachDefaultSystem (system: {
+      formatter = nixpkgs.legacyPackages.${system}.nixpkgs-fmt;
```
t184256, line 51: Nit: why do we export it?

t184256 (approved):
Approved in general.

flake.nix
Gerschtli: nix flake show won't show it, but it is necessary for nix fmt to work.

t184256: Huh, OK then.

t184256:
Don't forget to autosquash before merging.

PR #200, opened 2022-09-13
Bump bootstrap nix to 2.11.0

### LINKED ISSUE #199: Failed to turn on BTI protection
I'm unable to get a shell on Samsung galaxy z flip 4

```
/data/data/com.termux.nix/files/usr/bin/sh: /nix/store/9b6mgv7wzg964rf292s9x49n5g8yfgkp-glibc-2.32-54/lib/ld-linux-aarch64.so.1: failed to turn on BTI protection

[Process completed (code 127) - press Enter]
```

zhaofengli:
This PR bumps the bootstrap Nix version to 2.11.0. This brings in a newer version of glibc which fixes issues with devices with BTI functionality. Since newer Nix bootstrap tarballs no longer include coreutils, let's also remove our dependency on it.

Tested on the Exynos variant of Galaxy S22 (SM-S901B). Fixes #199.

### DIFF
--- a/modules/environment/login/login-inner.nix
+++ b/modules/environment/login/login-inner.nix
@@ -3,7 +3,7 @@
 { config, lib, customPkgs, writeText }:
 
 let
-  inherit (customPkgs.packageInfo) cacert coreutils nix;
+  inherit (customPkgs.packageInfo) cacert nix;
 in
 
 writeText "login-inner" ''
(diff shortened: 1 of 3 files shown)


t184256 (approved):
LGTM and big thank you for resolving this!

Not sure how avoiding coreutils relates to this issue, but I don't see downsides of doing that, so why not.

Tested to bootstrap and pass on-device self-tests on an Android 12L device.

t184256:
And the zipball goes down from 46M to 32M, impressive.

zhaofengli:
> Not sure how avoiding coreutils relates to this issue, but I don't see downsides of doing that, so why not.

The new bootstrap tarballs don't contain coreutils, so `PKG_COREUTILS` will be empty causing `${coreutils}/bin/mkdir` et al to just expand to `/bin/mkdir` which doesn't work.

t184256:
Understood, thank you.

Issue #180, opened 2022-05-11
Terminal font installation


PaulGrandperrin:
Because I cannot reopen issue #120 , I open a new one to continue the discussion.

[nix-darwin](https://github.com/LnL7/nix-darwin/blob/master/modules/fonts/default.nix?rgh-link-date=2022-04-30T21%3A50%3A39Z#L55) and [home-manager](https://github.com/nix-community/home-manager/blob/781d25b315def05cd7ede3765226c54216f0b1fe/modules/targets/darwin/fonts.nix#L14=) also need to copy fonts over filesystems boundaries and use the activation script to do it.

Let's do the same?

I can write a PR if I get the green light.

In the meantime, here what it looks like (when using `home-manager`): 
```nix
home.activation = {
  copyFont = let
      font_src = "${pkgs.nerdfonts.override { fonts = [ "FiraCode" ]; }}/share/fonts/truetype/NerdFonts/Fira Code Regular Nerd Font Complete Mono.ttf";
      font_dst = "${config.home.homeDirectory}/.termux/font.ttf";
    in lib.hm.dag.entryAfter ["writeBoundary"] ''
      ( test ! -e "${font_dst}" || test $(sha1sum "${font_src}"|cut -d' ' -f1 ) != $(sha1sum "${font_dst}" |cut -d' ' -f1)) && $DRY_RUN_CMD install $VERBOSE_ARG -D "${font_src}" "${font_dst}"
  '';
};
```

Other examples are available at #120

t184256:
I'm not against in principle, but probably will be of little help since my development machine suddenly decided to go to silicon heaven.

What's the characteristic execution time for such a snippet?

Gerschtli:
A simple option to specify a font would be a great addition. Would be worth creating a MR for that :)

Gerschtli:
This issue seems to be solved by #201. Feel free to reopen if you think otherwise :)

Issue #120, opened 2021-06-23
Is it possible to add fonts?


bbigras:
I'm using `fonts.fontconfig.enable = true;` and:

```nix
{
  home.packages = with pkgs; [
    meslo-lgs-nf
  ];
}
```

and I seem to be missing fonts with the powerlevel10k prompt.

t184256:
TL;DR: you can set one font

Fonts are hard to do right. I've started twice or thrice, never finished.

Font facts:

* in normal nix-on-droid usage (just running the terminal), none of the linux/home-manager font configuration facilities matter, fontconfig is ignored
* the only thing that renders fonts is [my fork of the Termux app](https://github.com/t184256/nix-on-droid-app)
* the fork has the styling addon baked in, so it looks for a `~/.termux/font.ttf`
* so, good news are, you can change the font
* bad news are, it has to be a regular file and not a symlink to `/nix/store` because, as far as the terminal app is concerned, there's no `/nix/store`. Auto-copying it on activation sounds hacky. Patching the app to do the sort of broken symlink resolution we already do for login script is definitely an option, but I'm basically allergic to Android development. PRs are welcome though.
* another bad thing is, Termux terminal emulator isn't exactly breathtaking with its font-rendering capabilities, even doing obliques instead of italics (my poor font)
* finally, there's no family detection, fallback chain or anything, none of this stuff. one ttf to rule them all

So, if you want a font with extra symbols, you'd have to find a TTF that has all the symbols you want in one file and then put it at `~/.termux/font.ttf`.

bbigras:
![Screenshot_20210624-121308.png](https://user-images.githubusercontent.com/24027/123297518-2a886000-d507-11eb-9bc1-ff2a267206d3.png)

It works! Thanks!

PaulGrandperrin:
@t184256 FYI, the people of  `nix-darwin` are doing this "hacky auto-copying on activation": https://github.com/LnL7/nix-darwin/blob/master/modules/fonts/default.nix#L55

I think the ease of use is so much worth this "hackyness".

Here's my code for those who want (using `home-manager`'s lib):
```nix
home.activation = {
  copyFont = lib.hm.dag.entryAfter ["writeBoundary"] ''
[...]

PR #201, opened 2022-09-18
modules/terminal: make font configurable

t184256:
Related: https://github.com/t184256/nix-on-droid-app/pull/2

@PaulGrandperrin  @Gerschtli, do you know why we haven't considered that and instead explored [copying the file over](https://github.com/t184256/nix-on-droid/issues/180)?

### DIFF
--- a/modules/terminal.nix
+++ b/modules/terminal.nix
@@ -1,0 +1,74 @@
+# Copyright (c) 2019-2022, see AUTHORS. Licensed under MIT License, see LICENSE.
+
+{ config, lib, ... }:
+
+with lib;
+
+let
+  cfg = config.terminal;
+in
+
+{
+
+  ###### interface
+
+  options = {
+
+    terminal.font = mkOption {
+      default = null;
+      type = types.nullOr types.path;
+      example = lib.literalExpression
+        ''"''${pkgs.terminus_font_ttf}/share/fonts/truetype/TerminusTTF.ttf"'';
+      description = ''
+        Font used for the terminal.
+        Must be a path to a TTF font.
+      '';
+    };
+
+  };
+
+
+  ###### implementation
+
+  config = {
+
+    build.activation =
+      let
+        fontPath =
+          if (lib.strings.hasPrefix "/nix" cfg.font)
+          then "${config.build.installationDir}/${cfg.font}"
+          else cfg.font;
+        configDir = "${config.user.home}/.termux";
+        fontTarget = "${configDir}/font.ttf";
+        fontBackup = "${configDir}/font.ttf.bak";
+      in
+      if (cfg.font != null) then
+        {
+          linkFont = ''
+            $DRY_RUN_CMD mkdir $VERBOSE_ARG -p ${configDir}
+            if [ -e ${fontTarget} ] && ! [ -L ${fontTarget} ]; then
+              $DRY_RUN_CMD mv $VERBOSE_ARG ${fontTarget} ${fontBackup}
+              $DRY_RUN_CMD echo "${fontTarget} has been moved to ${fontBackup}"
+            fi
+            $DRY_RUN_CMD ln $VERBOSE_ARG -sf ${fontPath} ${fontTarget}
+          '';
+        }
+      else
+        {
+          unlinkFont = ''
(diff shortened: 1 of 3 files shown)


PaulGrandperrin:
I think i never actually tried doing a symlink because of this:

https://github.com/t184256/nix-on-droid/issues/120#issuecomment-867219785

> * bad news are, it has to be a regular file and not a symlink to `/nix/store` because, as far as the terminal app is concerned, there's no `/nix/store`. Auto-copying it on activation sounds hacky. Patching the app to do the sort of broken symlink resolution we already do for login script is definitely an option, but I'm basically allergic to Android development. PRs are welcome though.

PaulGrandperrin:
Is this working?

t184256:
Seems like it is, yes =)

I guess I was wrong with my original analysis: yes, here I generate what's a broken symlink from nix-on-droid perspective, but it's a valid one from the outer point of view, and that's what matters ¯\\_(ツ)_/¯

Gerschtli:
Sounds legit, good idea, some minor things I found:

* Is the actual font path a gc root? Aka will it stay after a complete garbage collection?
* Should that really be `activationBefore`? Could be in normal `activation` I think.
* Should we care about keeping a backup/restore backup if there was a font installed manually before? (Maybe remove directory on unlink if empty)
* We should use `$VERBOSE_ARG` consistently in our activation scripts
* `isNull` is [deprecated](https://nixos.org/manual/nix/stable/language/builtins.html#builtins-isNull)

modules/terminal.nix
```diff
@@ -0,0 +1,52 @@
...
+{
+
+  ###### interface
+
+  options = {
+
+    terminal.font = mkOption {
+      default = null;
+      type = types.nullOr types.path;
+      example = "${pkgs.terminus_font_ttf}/share/fonts/truetype/TerminusTTF.ttf";
```
Gerschtli, line 20: That will render a more usable example.
```suggestion
      example = literalExpression ''"''${pkgs.terminus_font_ttf}/share/fonts/truetype/TerminusTTF.ttf"'';
```

Gerschtli:
Also could be worth adding to changelog :)

t184256:
Thanks for your review.

> Is the actual font path a gc root? Aka will it stay after a complete garbage collection?

It does, as the font path is embedded into the activation script.

> Should that really be activationBefore? Could be in normal activation I think.

Fixed.

> Should we care about keeping a backup/restore backup if there was a font installed manually before? (Maybe remove directory on unlink if empty)

Made it back up the font, but stopped at that. I like removing the directory even less.

> We should use $VERBOSE_ARG consistently in our activation scripts

Sprinkled.

> isNull is deprecated

Fixed.

> literalExpression

Accepted.

+ added support for both `/nix`-paths and non-mangled absolute paths.

Gerschtli (approved):
Cool, LGTM. But I am curious: Why don't you like removing empty directories created by nix-on-droid? home-manager behaves the same way.

Also, could be a good idea to log to the user, that you backuped a file.

t184256:
Updated.

One more important thing I forgot to ask: does the `terminal.font` interface look fine? Without a `terminal.enabled` knob.

t184256:
(will I ever get this pretty?..)

t184256:
How about this? (does nothing unless there was a link; then either restores backup or removes empty dir)

Gerschtli:
LGTM :)

t184256:
Thanks for your patience, squashing =)

MatrixManAtYrService:
Based on this conversation, it looks like this works.  Hurray!

I'm a bit of a noob though so like... how can I set the font?

t184256:
On January 13, 2023 11:39:02 PM GMT+01:00, Matt Rixman ***@***.***> wrote:
>Based on this conversation, it looks like this works.  Hurray!
>
>I'm a bit of a noob though so like... how can I set the font?
>
>

Short story: in file `~/.config/nixpkgs/nix-on-droid.nix` set option `terminal.font`. Example: https://t184256.github.io/nix-on-droid/nix-on-droid-options.html#opt-terminal.font

MatrixManAtYrService:
Thank you!

PR #202, opened 2022-09-20
login: Add fake /proc/uptime workaround for ps

### LINKED ISSUE #77: No /proc mount
Stuff like `ps` doesn't work, I'm assuming because /proc isn't mounted,
am I missing something?

zhaofengli:
This applies the workaround in https://github.com/SDRausty/TermuxArch/issues/175 when `/proc/uptime` is not readable. Fixes #77.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   set it to a TTF file to apply a font.
   An in-app `Styling` option will no longer work.
   Previously present file will be backed up to `~/.termux/font.ttf.bak`.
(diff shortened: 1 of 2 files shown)


Gerschtli:
Diff LGTM, need to test it. Could you add note in the changelog so everybody will know that `ps` finally works :)

Gerschtli:
Just thought about it: Could you write a small test just to ensure ps really works?

Gerschtli (approved):
Tested on device, works like a charm, thank you :)

Issue #156, opened 2021-12-14
Basic sshd module


Gerschtli:
I was tired of always copy-pasting the instructions from the Wiki to set up sshd, so I build this litte config:

```nix
let
  sshdTmpDirectory = "${config.user.home}/sshd-tmp";
  sshdDirectory = "${config.user.home}/sshd";
  pathToPubKey = "...";
  port = 8022;
in
{
  build.activation.sshd = ''
    $DRY_RUN_CMD mkdir $VERBOSE_ARG --parents "${config.user.home}/.ssh"
    $DRY_RUN_CMD cat ${pathToPubKey} > "${config.user.home}/.ssh/authorized_keys"

    if [[ ! -d "${sshdDirectory}" ]]; then
      $DRY_RUN_CMD rm $VERBOSE_ARG --recursive --force "${sshdTmpDirectory}"
      $DRY_RUN_CMD mkdir $VERBOSE_ARG --parents "${sshdTmpDirectory}"

      $VERBOSE_ECHO "Generating host keys..."
      $DRY_RUN_CMD ${pkgs.openssh}/bin/ssh-keygen -t rsa -b 4096 -f "${sshdTmpDirectory}/ssh_host_rsa_key" -N ""

      $VERBOSE_ECHO "Writing sshd_config..."
      $DRY_RUN_CMD echo -e "HostKey ${sshdDirectory}/ssh_host_rsa_key\nPort ${toString port}\n" > "${sshdTmpDirectory}/sshd_config"

      $DRY_RUN_CMD mv $VERBOSE_ARG "${sshdTmpDirectory}" "${sshdDirectory}"
    fi
  '';

  environment.packages = [
    (pkgs.writeScriptBin "sshd-start" ''
      #!${pkgs.runtimeShell}

      echo "Starting sshd in non-daemonized way on port ${toString port}"
      ${pkgs.openssh}/bin/sshd -f "${sshdDirectory}/sshd_config" -D
    '')
  ];
}
```

It is not really sophisticated and could be improved a lot but maybe it is worth adding that as a module?

Maybe it is just enough to keep it here or put it into the Wiki until we have a solid solution for how to run services in our nix-on-droid environment.

I mainly wanted to share that snippet in case anyone has the same problem :)

t184256:
Not to downplay the snippet's usefulness, but I'm afraid it's wiki material. Services starting up on activation is neither what I'd expect as a user nor what I'd want to maintain wrt interface stability and testing.

[...]

Issue #32, opened 2019-12-27
ssh server


mehlon:
It'd be nice to be able to access the Nix/Termux shell remotely. Can you add instructions how to do so?
It's very easy on Termux, but simply installing `nix-env -iA nixpkgs.openssh` then running `sshd` doesn't work out of the box.
https://wiki.termux.com/wiki/Remote_Access

t184256:
I didn't try this myself, but I recall someone reporting that it's possible:

https://logs.nix.samueldr.com/nixos-aarch64/2019-09-17
https://logs.nix.samueldr.com/nixos-aarch64/2019-09-18

Gerschtli:
For now you can follow these steps to create a working sshd server:

1. Install `openssh` via `nix-env` or `nix-on-droid` or what ever.
2. Create host key: `ssh-keygen -t rsa -b 4096 -f ssh_host_rsa_key -N ""`
3. Create `sshd_config` config file with following content:
    ```
    HostKey /absolute/path/to/ssh_host_rsa_key
    Port 8022
    ```
4. Create `~/.ssh/authorized_keys` and add a public key
5. Run sshd with absolute path like: `/data/data/com.termux.nix/files/home/.nix-profile/bin/sshd -f /path/to/sshd_config`
6. Connect to your phone via `ssh -p 8022 -l nix-on-droid -i /path/to/id_rsa <ip of phone>`

We will provide a module for the sshd config and service in the future to simplify the initial setup.

Kiwi:
That looks like what I did! I was going to do a write up on it but...  you know, life.

chenyulue:
I follows the @Gerschtli 's instructions, and finally run:
`ssh -p 8022 -l nix-on-droid -i ~/.ssh/id_rsa 192.168.0.101`

I can connect to my phone, but get the following strange results:

> -sh-4.4$ ls
> -sh: ls: command not found
> -sh-4.4$ echo $HOME
> /data/data/com.termux.nix/files/home
> -sh-4.4$ echo $PATH
> /usr/bin:/bin:/usr/sbin:/sbin:/nix/store/xx89cqfsn7f9ydcwmyiw0n110gwrhacv-openssh-7.9p1/bin
> -sh-4.4$ echo $PREFIX
> -sh-4.4$ cd /usr/bin
> -sh-4.4$ ls
> -sh: ls: command not found
> -sh-4.4$ nix-env -qa python
> -sh: nix-env: command not found
> -sh-4.4$

[...]

PR #203, opened 2022-09-21
modules/supervisord: init, modules/openssh: init

### LINKED ISSUE #54: Init daemon to manage all running processes
Hey, in the last weeks I was trying to build service-like modules, e.g. an sshd service that will execute some commands before starting like generating a host key and start the sshd daemon. To stop it, you have to kill the process, which needs to be done on exiting the app, if not it will run in the background.

Similar problem with `ssh-agent`, the agent will not be killed on exiting the app.

Of course one could define for these commands services and let login-inner stop all running services on app exit, but this is not really error-prone.

That is why I thought of something like an init daemon, so that all processes started will be killed, when killing the init daemon.

Thats all I have for now. What do you thing? Have you another/a better idea?

zhaofengli:
This PR adds support for running services with supervisord as well as an OpenSSH module. I wanted to run an SSH server (#32, #156) that gets automatically launched when Nix-on-Droid starts, complete with process supervision.

The `supervisord.programs.<name>` options are designed to resemble `systemd.services.<name>`:

```nix
{ pkgs, ... }:
{
  supervisord.programs.sshd = {
    path = [ pkgs.openssh ];
    autoRestart = true;
    script = ''
      ssh-keygen ...

      exec ${pkgs.openssh}/bin/sshd -D -f /etc/ssh/sshd_config
    '';
  };
}
```

Fixes #54.


modules/supervisord.nix
```diff
@@ -0,0 +1,177 @@
...
+        default = "/tmp/supervisor.log";
+      };
+      programs = lib.mkOption {
+        description = ''
+          Definition of supervisord programs.
+        '';
+        type = types.attrsOf programType;
+        default = {};
+      };
+      configFile = lib.mkOption {
```
Gerschtli, line 158: Why do you define an option for this? Do you have a use case for this option in mind? Otherwise, I would prefer to just use the `configFile` variable without this indirection.

modules/supervisord.nix
```diff
@@ -0,0 +1,177 @@
...
+      "supervisor.rpcinterface_factory" = "supervisor.rpcinterface:make_main_rpcinterface";
+    };
+  } // (lib.mapAttrs' (k: v: {
+    name = "program:${k}";
+    value = renderProgram v;
+  }) enabledPrograms);
+
+  configFile = format.generate "supervisord.conf" structuredConfig;
+
+  # Only expose the "supervisorctl" executable
```
Gerschtli, line 105: Out of interest: Why do you want to only expose this executable and not the others?

modules/supervisord.nix
```diff
@@ -0,0 +1,177 @@
...
+          Shell commands executed as the service's main process.
+        '';
+        type = types.lines;
+        default = "";
+      };
+      path = lib.mkOption {
+        description = ''
+          Packages added to the service's PATH environment variable.
+        '';
+        type = types.listOf (types.either types.package types.str);
```
Gerschtli, line 36: Why should we allow strings here?

modules/supervisord.nix
```diff
@@ -0,0 +1,177 @@
...
+        default = [];
+      };
+      autoRestart = lib.mkOption {
+        description = ''
+          Whether to automatically restart the process if it exits.
```
Gerschtli, line 41: Could you shortly describe what `unexpected` means? And/or provide a link to supervisord docs :)

modules/supervisord.nix
```diff
@@ -0,0 +1,177 @@
...
+        type = types.either types.bool (types.enum [ "false" "true" "unexpected" ]);
+        default = "unexpected";
+      };
+      environment = lib.mkOption {
+        description = ''
+          Environment variables passed to the service's process.
+        '';
+        type = types.attrsOf types.str;
+        default = {
+          PATH = lib.makeBinPath config.path;
```
Gerschtli, line 52: Defining PATH as default value can result in confusing behaviour. E.g. if I set `path` to a list with pkg1 and set anything in `environment`, pkg1 will not be available.

I would prefer setting PATH independent of `environment` and have an assertion to protect overrides like this.

Gerschtli (changes requested):
Looks cool, will continue the review later :)

modules/supervisord.nix
zhaofengli: I initially wanted to use the hardcoded config path to start supervisord itself in `login-inner`, but reloading is broken this way (what did I expect :stuck_out_tongue:). So this isn't really useful anymore and let's remove it.

zhaofengli (commented):
Thanks for the review! I'll squash the commits once everything is ready.

modules/supervisord.nix
```diff
@@ -0,0 +1,186 @@
...
+  options = {
+    supervisord = {
+      enable = lib.mkOption {
+        description = lib.mdDoc ''
+          Whether to enable the supervisord process control system.
+
+          This allows you to define long-running services in Nix-on-Droid.
+        '';
+        type = types.bool;
+        default = numPrograms != 0;
```
Gerschtli, line 127: Wouldn't it make more sense to check for enabled programs only?
```suggestion
        default = enabledPrograms != 0;
```

modules/supervisord.nix
```diff
@@ -0,0 +1,186 @@
...
+  });
+
+  renderAtom = val:
+    if builtins.isBool val then if val then "true" else "false"
+    else toString val;
+
+  renderProgram = program: let
+    section = {
+      inherit (program) command;
+      autorestart = program.autoRestart;
```
Gerschtli, line 79: I think it would also make sense to define `autostart` as an option, sounds like a common option.

modules/supervisord.nix
```diff
@@ -0,0 +1,186 @@
...
+    };
+
+    environment.packages = [ supervisorctl ];
+
+    build.activationAfter.reloadSupervisord = ''
+      if [ ! -e "${config.supervisord.socketPath}" ]; then
+        echo "Starting supervisord..."
+        ${cfg.package}/bin/supervisord -c /etc/supervisord.conf
+      else
+        ${cfg.package}/bin/supervisorctl -c /etc/supervisord.conf update
```
Gerschtli, line 182: A log message would be great. Also, please support dry runs (see `$DRY_RUN_CMD`)

modules/services/openssh.nix
```diff
@@ -0,0 +1,137 @@
+# Parts from nixpkgs/nixos/modules/services/networking/ssh/sshd.nix
+# MIT Licensed. Copyright (c) 2003-2022 Eelco Dolstra and the Nixpkgs/NixOS contributors
+
+{ pkgs, lib, config, ... }:
```
Gerschtli, line 4: Please have a look at the other files for examples of our licence header comment.

modules/services/openssh.nix
```diff
@@ -0,0 +1,137 @@
...
+        '';
+        type = types.listOf types.port;
+        default = [ 8022 ];
+      };
+      passwordAuthentication = lib.mkOption {
+        description = lib.mdDoc ''
+          Whether password authentication is allowed.
+        '';
+        type = types.bool;
+        default = true;
```
Gerschtli, line 71: Does not look like a reasonable default value. Why did you enable that by default? (Also I don't know of any way to set the password anyway :D)

modules/services/openssh.nix
```diff
@@ -0,0 +1,137 @@
...
+  config = lib.mkIf cfg.enable {
+    environment.etc = {
+      "ssh/sshd_config".source = sshdConf;
+      "ssh/moduli".source = "${cfg.package}/etc/ssh/moduli";
+    };
+
+    supervisord.programs.sshd = {
+      path = [ cfg.package ];
+      autoRestart = true;
+      script = ''
```
Gerschtli, line 115: https://github.com/NixOS/nixpkgs/blob/9b6e32ee67b4a556b4642f9fa1c91b9ebc7fc518/nixos/modules/services/networking/ssh/sshd.nix#L432-L434 sounds reasonable to also include here

Gerschtli (changes requested):
Please also mention the new modules in the changelog :) Would be great to have some tests for these new options (if feasible with reasonable effort).

Wanted to implement something like that since a very long time, I am glad you took the effort. Thank you a lot :)

Gerschtli:
For completion, it would be great to have native options to add ssh pub keys in `/etc/ssh/authorized_keys.d`. But could be done in a separate PR, however you like :)

Gerschtli:
Tested it on device: Running `supervisorctl stop sshd` does not stop the sshd process when there is an active ssh session.. Does not look like that would be expected behaviour.

modules/environment/login/login-inner.nix
```diff
@@ -16,6 +16,17 @@ writeText "login-inner" ''
     echo "If nothing works, open an issue at https://github.com/t184256/nix-on-droid/issues or try the rescue shell."
   fi
 
+  ${lib.optionalString config.supervisord.enable ''
+    set +e
+    if [ ! -e "${config.supervisord.socketPath}" ]; then
+      ${config.supervisord.package}/bin/supervisord -c /etc/supervisord.conf
```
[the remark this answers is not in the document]
zhaofengli: That makes sense to do, but I'm not sure what the cleanest way to do that is.

zhaofengli (commented):
> Tested it on device: Running `supervisorctl stop sshd` does not stop the sshd process when there is an active ssh session.. Does not look like that would be expected behaviour.

It's expected behavior because sshd forks a separate process to handle each connection. You can try by stopping sshd on NixOS (or any other distro): Existing sessions keep running but you can't establish new ones.

But yes, we do need to consider whether to want to terminate everything when all foreground Nix-on-Droid sessions exit. In regular Termux my manually started sshd processes still run when all sessions end.

modules/environment/login/login-inner.nix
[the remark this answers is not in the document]
t184256: I can see myself doing an 'Acquire wakelock', closing all shells and still expecting services to run, so not sure what's the ideal UX here. My vote is "notification is present <=> services are running".

What's the current situation?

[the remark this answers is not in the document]
Gerschtli: I think it is worth providing solutions to both use cases, so there could be an option to disable automatic shutdown after last session closes (I think using automatic shutdown as default sounds safer than the other way around).

But also not quite sure where such a shutdown could be implemented..

modules/supervisord.nix
```diff
@@ -0,0 +1,186 @@
...
+        type = types.bool;
+        default = true;
+      };
+      command = lib.mkOption {
+        description = lib.mdDoc ''
+          The command that will be run as the service's main process.
+        '';
+        type = types.str;
+      };
+      script = lib.mkOption {
```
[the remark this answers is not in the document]
Gerschtli: What do think about splitting set up logic from the actual command? I like the concept of ExecPreStart and similar lifecycle hooks of systemd that we could just implement for the user.

If you don't like it, please put a note about the necessity of `exec` in the description of `script` and clarify when each of the options are recommended to use and that not both options can be set simultaneously, ideally throwing an assertion error there to avoid unexpected behaviour.

modules/supervisord.nix
Gerschtli: Not really. For example I want to enable sshd (to be able to start and stop it on demand) but not enabling autostart

Gerschtli:
> It's expected behavior because sshd forks a separate process to handle each connection. You can try by stopping sshd on NixOS (or any other distro): Existing sessions keep running but you can't establish new ones.

Okay, but there is still a sshd process running even after the last ssh connection is closed when the sshd was stopped while session where open.

modules/services/openssh.nix
```diff
@@ -0,0 +1,130 @@
...
+    cat >$out <<EOL
+    ${uncheckedConf}
+    EOL
+
+    ssh-keygen -q -f mock-hostkey -N ""
+    sshd -t -f $out -h mock-hostkey
+  '';
+in {
+  options = {
+    services.openssh = {
```
Gerschtli, line 39: While playing around with this module on my device: I would like to set autostart to false and I would rather have an option like `services.openssh.autostart` than to use `supervisord.programs.sshd.autostart` or similar. Could you add an option like this?

modules/supervisord.nix
```diff
@@ -0,0 +1,189 @@
...
+          Environment variables passed to the service's process.
+        '';
+        type = types.attrsOf types.str;
+        default = {};
+      };
+      extraConfig = lib.mkOption {
+        description = lib.mdDoc ''
+          Extra structured configurations to add to the [program:x] section.
+        '';
+        type = types.attrsOf types.str;
```
Gerschtli, line 62: Because `renderAtom` supports also boolean, it would be reasonable to support booleans here too

modules/services/openssh.nix
```diff
@@ -0,0 +1,130 @@
...
+    };
+  };
+  config = lib.mkIf cfg.enable {
+    environment.etc = {
+      "ssh/sshd_config".source = sshdConf;
+      "ssh/moduli".source = "${cfg.package}/etc/ssh/moduli";
+    };
+
+    supervisord.programs.sshd = {
+      path = [ cfg.package ];
```
Gerschtli, line 103: This would not be necessary if you would use `ssh-keygen` with full path :)

modules/supervisord.nix
zhaofengli: Ok, added

modules/supervisord.nix
zhaofengli: Ok, added (both supervisord and sshd).

modules/supervisord.nix
```diff
@@ -94,7 +94,9 @@ let
       autorestart = program.autoRestart;
       environment = let
         # FIXME: Make more robust
-        escape = builtins.replaceStrings [ "%" ] [ "%%" ];
+        escape = s:
+          assert lib.assertMsg (!(lib.hasInfix "\"" s)) "supervisord.programs.<name>.environment: Values cannot have double quotes at the moment (${s})";
```
Gerschtli, line 98: Using asserts llike this will fail the whole evaluation and only show on assertion error at a time. Can you rather use NixOS style assertions? (https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/misc/assertions.nix)

modules/supervisord.nix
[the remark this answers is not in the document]
Gerschtli: But if you set e.g.
```sh
script = ''
  some-command
``;
```
signals like SIGTERM will not be forwarded by default to this process or is supervisord doing some magic there? These signals need to be forwarded to the actual main process.

Regarding lifecycle hooks: I agree that that this looks like an illusion but I am just thinking about improving UX through providing useful abstractions

modules/services/openssh.nix
Gerschtli: But then please be consistent and run `sshd` without absolute path. Mixing usage of PATH and absolute paths could result in bugs in the future when maintaining this script (because one could easily think that changing the PATH would be enough to use different sshd package. Consistency helps in the long run :)

zhaofengli:
There is another UX issue with the current approach. Currently `proot` waits for all traced processes to terminate, so the first Nix-on-Droid session which starts supervisord won't terminate after exiting the shell. If this isn't desirable, we can start supervisord in a separate proot invocation.

modules/environment/login/login-inner.nix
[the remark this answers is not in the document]
Gerschtli: We could replace `exec proot` in `/bin/login` with a fork and wait approach while redirecting process signals to the proot process, enabling us to run cleanup stuff after a session is closed. Counting of proot processes could then be used to shutdown supervisord. I can not think of a downside of this approach (and I don't see an alternative).

Gerschtli:
Why not detach the process from the invoking session with e.g. disown?

zhaofengli:
> Why not detach the process from the invoking session with e.g. disown?

Sadly this can't work because proot needs to ptrace all processes in the session to perform its hooking, so it has to be there as long as any child processes exist. There doesn't appear to be an option to have proot itself fork into the background if the "main process" exits.

t184256:
Random idea, feel free to disregard:

```
login -> proot -> login-inner -> session
       \
        -> (supervisord overseeing process) -> proot -> login-inner -> supervisord
```

zhaofengli:
> Random idea, feel free to disregard:
> 
> ```
> login -> proot -> login-inner -> session
>        \
>         -> (supervisord overseeing process) -> proot -> login-inner -> supervisord
> ```

That's what I meant by "start supervisord in a separate proot invocation." If this is desired I'll implement it later this week.

t184256:
I still somehow thought that meant starting it from proot with extra precautions. Sorry for the noise then.

As for desireability, I'm not sure. Sounds rather involved to me wrt starting/stopping it (via signals?). If you figure out some alternative that simplifies things at the expense of, say, some support from the app, I'm open for discussing it further.

Gerschtli:
Hey @zhaofengli , is there anything where I could support you with this PR? :)

zhaofengli:
Sorry, got distracted and forgot about this :/ Termux does appear to have a special type of session for background tasks:

Termux notification with background tasks active

One integration is implemented in Termux:Widgets, where you can drop a shell script in `~/.shortcuts/tasks` and add a shortcut that launches the script in the background. We should look into whether it's possible to start such a task with `/system/bin/am start` or similar outside PRoot. If so, then things should be cleaner with the user still having a persistent notification even with no interactive sessions active.

diamondburned:
Is there a way to test this PR without rebuilding the fork? Maybe a way to patch these modules into the config?

Gerschtli:
Yes, either use @zhaofengli's fork as channel/flake input or create your own fork of nix-on-droid and rebase his branch onto latest master to keep all changes that were merged in the meantime. No need for a new bootstrap zip because there were no changes :)

diamondburned:
> create your own fork of nix-on-droid and rebase his branch onto latest master to keep all changes that were merged in the meantime

Sorry for being off-topic; how would I do this, exactly? I'm currently not using Flakes. Do I swap the Nix channel with one that points to this commit?

Gerschtli:
nix-channel --list shows you the current nix-on-droid url. That's a url with repo owner, repo name and rev (probably master or release-22.11). Replace owner and rev and set within .nix-channels, afterwards run nix-channel --update and nix-on-droid switch :)

modules/services/openssh.nix
```diff
@@ -0,0 +1,141 @@
...
+                ${if k ? bits then "-b ${toString k.bits}" else ""} \
+                ${if k ? rounds then "-a ${toString k.rounds}" else ""} \
+                ${if k ? comment then "-C '${k.comment}'" else ""} \
+                ${if k ? openSSHFormat && k.openSSHFormat then "-o" else ""} \
+                -f "${k.path}" \
+                -N ""
+          fi
+        '')}
+
+        exec ${cfg.package}/bin/sshd -D -f /etc/ssh/sshd_config
```
Shados, line 137: Should probably add the `-e` flag so log output goes to stderr, which supervisord can actually pick up (unlike syslog).

expenses:
What's the status of this PR? I've been using it for some basic services and I'm pretty happy with it. Would be good to get it merged if possible.

bbigras:
Another friendly ping. Any progress on this?

modules/environment/login/login-inner.nix
```diff
@@ -16,6 +16,12 @@ writeText "login-inner" ''
     echo "If nothing works, open an issue at https://github.com/t184256/nix-on-droid/issues or try the rescue shell."
   fi
 
+  ${lib.optionalString config.supervisord.enable ''
+    (if [ ! -e "${config.supervisord.socketPath}" ]; then
```
Marlinc, line 20: This check fails when the supervisord socket file still exists but supervisor isn't running. I'm not sure what a good alternative would be.

Shados: `nc -zU ${config.supervisord.socketPath}` will fail if the socket file isn't actually connected, and succeed without actually sending anything on the socket otherwise.

non-bin:
Bump

PR #205, opened 2022-09-26
modules/build/activation: Prepend to PATH instead of setting it outright

Shados:
Prior to this change: If your nix.conf has `max-jobs = 0` and `builders = ssh://...`, thus configuring it to do all build jobs remotely, activation would fail due to `nix-env` being unable to find `ssh` in `PATH`. By prepending to `PATH` instead of overriding it entirely, this is avoided.

Of course, `nix-env` probably shouldn't need a remote builder in order to simply upgrade to an already-built derivation, but that's a Nix issue if anything.

### DIFF
--- a/modules/build/activation.nix
+++ b/modules/build/activation.nix
@@ -35,7 +35,7 @@ let
     set -eu
     set -o pipefail
 
-    export PATH="${activationBinPaths}"
+    export PATH="${activationBinPaths}:''${PATH:+:''${PATH}}"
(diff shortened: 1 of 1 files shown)


Gerschtli (changes requested):
This change will result in potential harder to debug problems and non reproducible builds because the activation script has now undefined runtime dependencies. I would rather see an approach like home-manager's `home.extraActivationPath` to add more packages into paths if necessary. Then at least the activation script works independently of the environment.

Shados:
Fair, that's a better approach. I'll close this for now.

PR #208, opened 2022-10-11
home-manager: fix directory not found error

Gerschtli:
Prepend of /. produces directory not found error when building on a non nix-on-droid environment. The etc module works also with just a string.


t184256 (approved):
I'm always for removing code =)

PR #204, opened 2022-09-24
Flakify everything

Gerschtli:
Marked as draft because it is based on #196.

Basically flakified bootstrap build process and fakedroid.

Related: https://github.com/t184256/nix-on-droid/issues/148, https://github.com/t184256/nix-on-droid/projects/2

Issue #36, opened 2020-01-04
Visualisation of available options


Gerschtli:
I would like to have a auto-generated list of all available options, similar to home-manager's [html page](https://rycee.gitlab.io/home-manager/options.html) or as output of `man nix-on-droid`.

I'm open for ideas where and how to implement this.

Gerschtli:
Was able to setup docs for an other project, so should not be that complicated to apply the same things here. Will do as soon as I have more time.
See https://github.com/Gerschtli/nix-formatter-pack/commit/66cec9b45b1ea2bc9e33422ee0f7dd18f6717dc6 and https://github.com/Gerschtli/nix-formatter-pack/commit/0fb492cd802b2b0b4bfe2f0035b68f988fb21cff for CI/deployment.

Gerschtli:
Solved here: https://t184256.github.io/nix-on-droid/nix-on-droid-options.html

PR #211, opened 2022-10-26
Add html and man pages for available options

Gerschtli:
Adds rendered list of available options. We could potentially thinks about moving github wiki, readme and changelog documentation to these html pages. 

For local testing, run 
```sh
nix build ".#manualHtml"
open result/share/doc/nix-on-droid/index.html

nix build ".#manPages"
man --local-file result/share/man/man5/nix-on-droid-options.5.gz
```

This PR contains also a new github workflow to automatically deploy latest docs to gh pages on every master push.

Fixed some option descriptions for correct formatting and renamed `Nix-on-Droid` occurrences to `nix-on-droid` for consistency.

Interesting, potential follow-up tasks:
- man page for nix-on-droid
- a `documentation.manPages.enable` option to install the man page on device.

Requires https://github.com/t184256/nix-on-droid/pull/204.
Solves https://github.com/t184256/nix-on-droid/issues/36.


Gerschtli:
Force push was only rebase on master

Gerschtli:
Oh cool, there is a "Compare"  button to see what changed between force-pushes. Not sure if I never saw it or it was added recently, but then we can just autosquash before pushing and still be able to see the differences :)

docs/man-pages.xml
```diff
@@ -0,0 +1,36 @@
+<!-- Copyright (c) 2019-2022, see AUTHORS. Licensed under MIT License, see LICENSE. -->
+<reference xmlns="http://docbook.org/ns/docbook"
+           xmlns:xlink="http://www.w3.org/1999/xlink"
+           xmlns:xi="http://www.w3.org/2001/XInclude">
+ <title>Nix-on-Droid Reference Pages</title>
```
t184256: is it going anywhere?

t184256 (commented):
Casing: should Android app also lowercase it now? IMO this would look rather odd on Android. (suggestion: lowercase when monospace, heh)

Extra entry:
```
       _module.args

           Type: lazy attribute set of raw value

           Declared by:
               <nix-on-droid/lib/modules.nix>
```
can this be hidden?

Gerschtli:
> Casing: should Android app also lowercase it now? IMO this would look rather odd on Android. (suggestion: lowercase when monospace, heh)

Good point. We can try to do that :D

> _module.args

I tried to hide it, didn't manage to do so. Basically, it is an option you can use, so it is not completely wrong :D Also home-manager docs has (obviously, because of nmd) the same behaviour.

docs/man-pages.xml
Gerschtli: But as mentioned in the PR description
> a `documentation.manPages.enable` option to install the man page on device.

We could install it on the device for easy access to option docs.

Gerschtli:
Tried to find a consistent way for the casing, what do you think?

t184256 (approved):

PR #212, opened 2022-10-26
Add nix-formatter-pack

Gerschtli:
I build a nice little tool to easily setup multiple formatters (we currently use nixpkgs-fmt and statix, this adds [deadnix](https://github.com/astro/deadnix)). The benefit is that running `nix fmt` fixes all issues found by all tools and `nix flake check` checks for all tools.

This is very opinionated so feel free to close it, if you don't like it :)

Requires https://github.com/t184256/nix-on-droid/pull/204.

### DIFF
--- a/.github/workflows/lints.yml
+++ b/.github/workflows/lints.yml
@@ -3,7 +3,7 @@ on:
   pull_request:
   push:
 jobs:
-  cachix:
+  lint:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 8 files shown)


t184256 (approved):
Thanks. The accompanying changes seem reasonable, so let's try it.

Gerschtli:
rebased on master. Thanks for the review!

PR #223, opened 2022-11-18
environment.path: fix installation with nix profile

Gerschtli:
Manual removal and installation as two non-atomical steps is required because of https://github.com/NixOS/nix/issues/6349.

Related https://github.com/nix-community/home-manager/pull/3367

### DIFF
--- a/modules/environment/path.nix
+++ b/modules/environment/path.nix
@@ -38,7 +38,19 @@ in
 
     build.activation.installPackages = ''
       if [[ -e "${config.user.home}/.nix-profile/manifest.json" ]]; then
-        $DRY_RUN_CMD nix profile install ${cfg.path}
+        # manual removal and installation as two non-atomical steps is required
+        # because of https://github.com/NixOS/nix/issues/6349
+
+        nix_previous="$(command -v nix)"
+
+        nix profile list \
+          | grep 'nix-on-droid-path$' \
+          | cut -d ' ' -f 4 \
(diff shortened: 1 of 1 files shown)


t184256:
Wait, so, it used to accumulate packages before without removing? Or am I misunderstanding something? Do we want a test for this?

Gerschtli:
In case of no flake input/channel updates, yes. In case of updates, the activation would have failed because there would have been an error like "/bin/abc is already installed, not able to install it in the profile" like shown in the related HM issue.

Sure, good idea, will add a test.

Gerschtli:
Regarding the test: Not sure how to do this properly because this means I need to convert the default profile to "nix profile"-type and can not really roll it back to "nix-env"-type

```
error: profile '/nix/var/nix/profiles/per-user/nix-on-droid/profile' is incompatible with 'nix-env'; please use 'nix profile' instead
```

So we can either test `nix-env` or `nix profile` activation logic but not both. Maybe we need to rethink our tests a little bit, also with the fact that they require a channel-setup and don't work with flakes..

Gerschtli:
@t184256 How do we want to go on here? I think it would be great to have this fix before creating 21.11 release.

Do you have suggestions how to write a test for that?

t184256:
As rethink our tests to nuke environment sounds overly grandiose, maybe we can merge that with just the manual testing.

Gerschtli:
Created https://github.com/t184256/nix-on-droid/issues/233 to don't forget it.

I tested it manually on device.

t184256 (approved):

PR #227, opened 2022-11-24
Export pkgs

573:
See @Gerschtli's suggestion in https://github.com/t184256/nix-on-droid/discussions/226#discussioncomment-4220726


Gerschtli (approved):
Thank you!

PR #229, opened 2022-11-25
fix(terminal.nix): quote font paths

reo101:
**Problem:**

Currently, if one wants to use a font, whose path contains spaces, for instance a nerdfont:

```nix
{
  terminal.font =
    let
      firacode = pkgs.nerdfonts.override {
        fonts = [ "FiraCode" ];
      };
      fontPath = "share/fonts/truetype/NerdFonts/Fira Code Regular Nerd Font Complete Mono.ttf";
    in
      "${firacode}/${fontPath}";
}
```

`nix-on-droid` would fail to create a correct symlink because the references to `${fontTarget}` aren't quoted which leads to `ln` treating the space-separated parts of the font name as different arguments.

Quoting `man ln`/`LN(1)`:

```
     Given more than two arguments, ln makes links in target_dir to all the named
     source files.  The links made will have the same name as the files being linked
     to.
```

which isn't the wanted behaviour here.

**Solution:**

Quote all unquoted references to nix variables in the `linkFont` script in `modules/terminal.nix`

**Demo:**

In a `nix-on-droid` configuration using my fork `github:reo101/nix-on-droid/reo101/font_path_quoting` the symbolink link is created correctly:

image

### DIFF
--- a/modules/terminal.nix
+++ b/modules/terminal.nix
@@ -44,24 +44,24 @@ in
       if (cfg.font != null) then
         {
           linkFont = ''
-            $DRY_RUN_CMD mkdir $VERBOSE_ARG -p ${configDir}
-            if [ -e ${fontTarget} ] && ! [ -L ${fontTarget} ]; then
-              $DRY_RUN_CMD mv $VERBOSE_ARG ${fontTarget} ${fontBackup}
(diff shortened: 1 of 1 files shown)


Gerschtli (approved):
Thank you very much, great PR description and good catch!

PR #232, opened 2022-11-30
deploy: allow specifying urls not ending with source.tar.gz

t184256:
For example, the official bootstrap ones.


Gerschtli (approved):
LGTM :)

Issue #230, opened 2022-11-27
docs: terminal.font is allowed to be an OTF file too


ajgrf:
The docs for `terminal.font` currently say that it “must be a path to a TTF font,” however OTF files work fine as well (the link name must be `font.ttf` regardless). TTC files also appear to work, but because TTC files are basically a bunch of TTFs wrapped up into a single file, there’s no way to control which typeface gets used.

So almost every font package in nixpkgs can be used by nix-on-droid. I think it would be good to reflect this in the docs.

Gerschtli:
Fixed with #234

PR #234, opened 2022-12-01
22.11

Gerschtli:
PR to prepare everything for 22.11.

* The state version is a sensible value that should explicitly set in the users config. Following home-manager: https://github.com/nix-community/home-manager/pull/3043.
* Updating qemu, nix and pinned nixpkgs. Store path to proot updated because not building, see below
* Updating references of 22.05 to 22.11
* Add stateVersion 23.05 (because master after release-22.11 will essentially be 23.05 (also following home-manager's approach)

Currently, bootstrap is not building:

```
$> nix build .#bootstrapZip --impure -L
nix-directory> unpacking sources
nix-directory> unpacking source archive /nix/store/4hk7sv1pdp8ip0ssnk88ri91fd2d9k4p-nix-2.11.1-aarch64-linux.tar.xz
nix-directory> source root is nix-2.11.1-aarch64-linux
nix-directory> setting SOURCE_DATE_EPOCH to timestamp 1663257321 of file nix-2.11.1-aarch64-linux/install-systemd-multi-user.sh
nix-directory> patching sources
nix-directory> configuring
nix-directory> no configure script, doing nothing
nix-directory> building
nix-directory> terminate called after throwing an instance of 'nix::SysError'
nix-directory>   what():  error: reading symbolic link '/proc/self/exe': No such file or directory
error: builder for '/nix/store/5qnqr259jbqsq95lfwc7ln0nh5lb407i-nix-directory.drv' failed with exit code 255;
```

Issue seems to be on `nix-store --init` after nixpkgs update, see https://github.com/t184256/nix-on-droid/blob/master/pkgs/nix-directory.nix#L46

### DIFF
--- a/modules/version.nix
+++ b/modules/version.nix
@@ -12,7 +12,6 @@ with lib;
 
     system.stateVersion = mkOption {
       type = types.enum [ "19.09" "20.03" "20.09" "21.05" "21.11" "22.05" ];
-      default = "19.09";
       description = ''
         It is occasionally necessary for Nix-on-Droid to change
         configuration defaults in a way that is incompatible with
--- a/modules/version.nix
+++ b/modules/version.nix
@@ -11,7 +11,16 @@ with lib;
   options = {
 
     system.stateVersion = mkOption {
-      type = types.enum [ "19.09" "20.03" "20.09" "21.05" "21.11" "22.05" ];
+      type = types.enum [
+        "19.09"
+        "20.03"
+        "20.09"
+        "21.05"
+        "21.11"
+        "22.05"
+        "22.11"
+        "23.05"
+      ];
       description = ''
         It is occasionally necessary for Nix-on-Droid to change
         configuration defaults in a way that is incompatible with
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -35,6 +35,8 @@ let
         pkgs = pkgs.lib.mkForce pkgs; # to override ./modules/nixpkgs/config.nix
       };
 
+      system.stateVersion = "22.05";
+
       # Fix invoking bash after initial build.
       user.shell = "${initialPackageInfo.bash}/bin/bash";
 
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -35,7 +35,7 @@ let
         pkgs = pkgs.lib.mkForce pkgs; # to override ./modules/nixpkgs/config.nix
       };
 
-      system.stateVersion = "22.05";
+      system.stateVersion = "22.11";
 
       # Fix invoking bash after initial build.
       user.shell = "${initialPackageInfo.bash}/bin/bash";
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 ### New Options
 
 * Terminal font now should be specified using `terminal.font` option,
-  set it to a TTF file to apply a font.
+  set it to any file containing a font to apply it.
   An in-app `Styling` option will no longer work.
   Previously present file will be backed up to `~/.termux/font.ttf.bak`.
 * Add option `environment.motd` to edit the startup message that is printed in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@
   name
 * Add html and man pages with all available options, see <https://t184256.github.io/nix-on-droid/>
   and `nix build github:t184256/nix-on-droid#manPages`
+* The option `system.stateVersion` does not have a default value anymore.
+  Previously, its default was `"19.09"` the release the state version was
+  introduced.
 
 ## Release 22.05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,29 @@
 
 ## Release 22.11
 
+### Compatibility considerations
+
+* Fix usage of `extraSpecialArgs`: All values were previously set in
+  `_module.args` instead of passing them as `specialArgs` into `evalModules`.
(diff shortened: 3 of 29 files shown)


Gerschtli:
Found a solution:
* mounting `/proc` helps, not sure about other side effects
* needed to create the nix state directory (`mkdir --parents ${buildRootDirectory}/nix/var/nix/db`) in order to tell nix to use the manually created `/nix` directory instead of a chroot nix store in `/homeless-shelter/.local/share/nix/root`. related warning that is printed without this fix:
  ```
  warning: '/nix' does not exist, so Nix will use '/homeless-shelter/.local/share/nix/root' as a chroot store
  ```

Gerschtli:
BTW there are quite some old branches that are merged and/or maybe not needed anymore. Except for current open PR branches, your `testing` branch and the `release-*` branches (and maybe `pre-module-system` branch), are the other branches still required?
Just for some house-keeping :)

```
  remotes/origin/FAILED-on-device-proot-building a8c3881 Attempt to fix proot static compilation
  remotes/origin/deploy-fix                      1194ecf deploy: allow specifying urls not ending with source.tar.gz
  remotes/origin/does-this-fix-make              d627071 pkgs/cross-compiling/proot: fix seccomp+setresxid
  remotes/origin/fix-flake-prompt                b623074 login-inner: fix default choice on bootstrap flakes prompt
  remotes/origin/fix-profile                     048b22f environment.path: fix installation with nix profile
  remotes/origin/master                          4e58ce7 fix(terminal.nix): quote font paths
  remotes/origin/pre-module-system               a505862 Add less to basic-environment, nix needs a $PAGER.
  remotes/origin/prerelease-20.09                fed9bbe pkgs/qemu-aarch64-static: update
  remotes/origin/prerelease-22.05                b286ac1 pkgs/nix-directory: update nix
  remotes/origin/prerelease-22.05-pt2            0ad6321 pkgs: use non-prebuilt SDK
  remotes/origin/prerelease-22.05-unmerged-work  d0b08d2 proot-termux: simplify module
  remotes/origin/prerelease-22.11                8816289 bootstrap: update nixpkgs version
  remotes/origin/release-19.09                   d551f63 Add a comment to clarify setting user.shell
  remotes/origin/release-20.03                   b651458 Fix Cachix in Travis with `trusted-users = travis`
  remotes/origin/release-20.09                   0dec5d4 modules/user: use runCommandLocal for idDerivation
  remotes/origin/release-21.05                   3e3a4e2 overlays/lib/nixpkgs.nix: use system
  remotes/origin/release-21.11                   e87f42e modules/upgrade: fix missing /dev/shm on upgrade from <21.11
  remotes/origin/release-22.05                   d49fd3a CHANGELOG: Release 22.05
  remotes/origin/testing                         62cf934 fixup ssh: don't write to stdout
  remotes/origin/update-sdk                      dda3bd3 WIP: compile proot dynamically with non-prebuilt SDK
  remotes/origin/wip-non-static-proot            07a89d7 WIP non-static proot
```

modules/version.nix
```diff
@@ -11,8 +11,16 @@ with lib;
...
-      default = "19.09";
+      type = types.enum [
+        "19.09"
+        "20.03"
+        "20.09"
+        "21.05"
+        "21.11"
+        "22.05"
+        "22.11"
+        "23.05"
```
SuperSandro2000, line 22: I would recommend to make this option not an enum. This makes updated in the future easier and frankly out should not matter of I set this to 22.12.

home-manager is doing the same and it is annoying with every release and only created extra work.

pkgs/default.nix
```diff
@@ -35,6 +35,8 @@ let
         pkgs = pkgs.lib.mkForce pkgs; # to override ./modules/nixpkgs/config.nix
       };
 
+      system.stateVersion = "22.11";
```
SuperSandro2000, line 38: Hard coding a state version inside here forces users to use mkForce which doesn't be required. I would recommend to either use lib.mkDefault or better omit it and force users to set it themselves.

SuperSandro2000 (commented):
I would also recommend to use more variables for the nixos version for the next update.

modules/version.nix
Gerschtli: Why should it be possible to allow arbitrary values here? Setting stateVersion ="foo" doesn't make any sense and is prone to errors because we can not use this value at all.

pkgs/default.nix
Gerschtli: This change only affects the bootstrap zip generation. For this we create a nix-on-droid configuration. On device, this file here is not used anymore.

Gerschtli:
> I would also recommend to use more variables for the nixos version for the next update.

What do you mean by that? What kind of variables?

t184256 (approved):
Thanks for the time and effort, I don't see anything to complain about!

Manually tested:

* basic bootstrap
* clean default upgrade
* upgrade with a home-manager-enabled setup

Please rebase, maybe also pick up #230 since it's nearly almost a changelog issues.

t184256:
> BTW there are quite some old branches that are merged and/or maybe not needed anymore.

Didn't know I had so many failed attempts to update SDK and build proot differently =/ Cleaned up some of the branches.

Gerschtli:
Rebased and added a commit for #230. Also dropped the workaround for default nix version < 2.4. If that is fine, you can merge and do the usual release steps :)

modules/version.nix
SuperSandro2000: Any value that matches the regex ``\d\d\.\d\d`` should work.

PR #270, opened 2023-07-02
modules/nix: use structural `settings` (Nix RFC 42)

ShamrockLee:
This patch ports the structural settings from nixos/nix-daemon and adjust the description based on those from previous options and the Home Manager nix.settings option.

If applied, users could specify their `nix.conf` configuration in a structural way, using attribute names identical to the official documentation of Nix, and merge the corresponding attributes from different modules. We don't have to add or change any of the attributes manually when configuration options are added / changed upstream, and the specified configurations are automatically checked using the specified Nix package.


modules/environment/nix.nix
```diff
@@ -2,17 +2,94 @@
...
 let
   cfg = config.nix;
+
   renameNixOpt = old: new:
     (mkRenamedOptionModule [ "nix" old ] [ "nix" new ]);
+
+  isNixAtLeast = versionAtLeast (getVersion cfg.package);
+
+  nixConf =
+    assert isNixAtLeast "2.2";
```
t184256, line 24: Even `nixVersions.minimum` is now 2.3, I don't think we need that.

modules/environment/nix.nix
```diff
@@ -2,17 +2,94 @@
...
+    let
+
+      mkValueString = v:
```
t184256, line 27: Can we reuse the nixpkgs version here?

t184256 (changes requested):
Thanks for the PR. I'd like to apologize, I have limited availability for the first half of July, and so far I could only give it a superficial review. I do remember about it and I will return to it in a while.

t184256:
Seems to function fine, please address the comments from the previous review.

ShamrockLee:
Sorry, for late respond. I'm working on an oral presentation for a conference about Nix, and will not be responsive until August.

t184256:
Oh, sure, no rush, take your time (and consider sharing the talk =).

t184256:
Are you still interested in mainlining this?

ShamrockLee:
Oh! I almost forgot about this one.

ShamrockLee:
Since then, things like the Nix settings and Nix flake registries have been splitted from `nix-daemon.nix` to their own module file (`nix.nix` and `nix-flakes.nix`). It would be great to reuse them.

ShamrockLee:
#329 demonstrate the possible strategy to reuse the upstream macro files.

PR #275, opened 2023-07-13
Add possibility to use fakedroid from outside the repo

### LINKED ISSUE #274: Way to run fakedroid from users's flake
Currently fakedroid is supposed to be run from the clone of `nix-on-droid` repository.

I tried to add [fakedroid](https://github.com/t184256/nix-on-droid/blob/2301e01d48c90b60751005317de7a84a51a87eb6/flake.nix#L67) to outputs of my flake with my system configs but it obviously didn't work - fakedroid expects to have some scripts in the repo.

Th reason I want it is to experiment quickly with configs without uploading and rebuilding them to my phone - it would be handy to have a single command that will create a VM (or sandbox) with my custom configs, not default ones.

Nixos-mobile has this with their `config.mobile.outputs.uefi.vm` package.

Gerschtli:
Resolves #274

### DIFF
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -38,4 +38,5 @@ pkgs.runCommand
       --subst-var-by prootTest "${bootstrap.customPkgs.prootTermuxTest}" \
       --subst-var-by installationDir "${bootstrap.config.build.installationDir}" \
       --subst-var-by homeDir "${bootstrap.config.user.home}" \
+      --subst-var-by flakeRootDir "${toString ../.}"
   ''
--- a/README.md
+++ b/README.md
@@ -239,6 +239,12 @@ with the command as arguments:
 nix run --impure ".#fakedroid" -- ls -l
 ```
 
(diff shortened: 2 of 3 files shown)


tests/default.nix
```diff
@@ -38,4 +38,5 @@ pkgs.runCommand
       --subst-var-by prootTest "${bootstrap.customPkgs.prootTermuxTest}" \
       --subst-var-by installationDir "${bootstrap.config.build.installationDir}" \
       --subst-var-by homeDir "${bootstrap.config.user.home}" \
+      --subst-var-by flakeRootDir "${toString ../.}"
```
t184256, line 41: Will that make fakedroid ignore uncommitted changes and always build environment from the commit it itself is built from?

Gerschtli: It will build with all changes tracked by git. Behaves the same as every other interaction with flakes in git repositories.

t184256:
Seems to work OK. Not sure I'm a fan of it always using the repo/commit it's built from, hope it won't bring confusion.

PR #277, opened 2023-07-20
Add networking options.

kurnevsky:
Tested on my phone with yggdrasil - works fine :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Release 23.11 (unreleased)
 
+### New Options
+
+* New options `networking.hosts`, `networking.hostFiles` and
+  `networking.extraHosts` for `/etc/hosts` configuration.
+
 ## Release 23.05
 
 ### New Options
--- a/tests/on-device/config-flake-hosts.cfg.nix
+++ b/tests/on-device/config-flake-hosts.cfg.nix
@@ -1,0 +1,19 @@
+{ pkgs, ... }:
+
+{
+  system.stateVersion = "23.05";
+
+  networking = {
+    hosts."127.0.0.2" = [ "a" "b" ];
+
+    extraHosts = ''
+      127.0.0.3 c
+    '';
+
+    hostFiles = [
+      (pkgs.writeText "hosts" ''
+        127.0.0.4 d
+      '')
+    ];
+  };
+}
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,3 +1,4 @@
 Alexander Sosedkin <monk@unboiled.info>
 Tobias Happ <tobias.happ@gmx.de>
 Bruno Bigras <bigras.bruno@gmail.com>
+Evgeny Kurnevsky <kurnevsky@gmail.com>
--- a/modules/environment/networking.nix
+++ b/modules/environment/networking.nix
@@ -1,19 +1,90 @@
-# Copyright (c) 2019-2021, see AUTHORS. Licensed under MIT License, see LICENSE.
+# Copyright (c) 2019-2023, see AUTHORS. Licensed under MIT License, see LICENSE.
+
+# Inspired by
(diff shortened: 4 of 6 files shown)


kurnevsky:
Hm, actually I might try to replicate more options from https://github.com/NixOS/nixpkgs/blob/nixos-23.05/nixos/modules/config/networking.nix

kurnevsky:
ok, seems to work. I don't really know why nixpkgs [defines](https://github.com/NixOS/nixpkgs/blob/nixos-23.05/nixos/modules/config/networking.nix#L169) 127.0.0.2 as localhost, so I didn't do that :)

kurnevsky:
> 127.0.0.2 as localhost

It's something from [nss-myhostname](https://www.freedesktop.org/software/systemd/man/nss-myhostname.html), but we don't have systemd so we don't really need it.

kurnevsky:
Ping :)

t184256:
Sorry for a slow response.

1. If you're borrowing from nixpkgs, you need to add attribution. grep for 'under MIT license as well' for examples.
2. Please bump the copyright year.
3. I'm not a fan of the overdone nixos interface, but, I guess, we shouldn't deviate.
4. But if we stick to it, I'd like a testcase exercising all these options + assertion path. You can write one or I can write one, but it might take some time for me to find time for that.

kurnevsky:
> Sorry for a slow response.

No problem. Just wanted to make sure it's not forgotten.

> I'm not a fan of the overdone nixos interface, but, I guess, we shouldn't deviate.

I personally use only `networking.hosts`, so I'm not really sure how other settings are useful. There are advantages of keeping it in sync with nixpkgs, but it's more complicated. I'm fine either way, I guess :)

> But if we stick to it, I'd like a testcase exercising all these options + assertion path. You can write one or I can write one, but it might take some time for me to find time for that.

I can try to do that. It's me who wants these settings after all :)

kurnevsky:
> But if we stick to it

So should we? :)

kurnevsky:
ok, added a test. Wanted to use https://github.com/bats-core/bats-assert#assert_line but couldn't figure out how to add it in shebang - I guess some more complex changes are required for it.

t184256 (commented):
> added a test

good. hope `grep` is available in tests.

> btw, should I add myself to authors now? :)

Feel free to. My idea was to let anyone edit it when they submit a PR and think it's substantial enough. BTW, if you know where to express this idea so that all contributors get it, I'm all ears.

Also, consider adding a changelog entry.

kurnevsky:
> good. hope grep is available in tests.

It is.

> BTW, if you know where to express this idea so that all contributors get it, I'm all ears.

Well, readme with  or  `Contributing.md` with some basic steps should be enough :)

> Also, consider adding a changelog entry.

Done.

tests/on-device/config-flake-hosts.cfg.nix
```diff
@@ -0,0 +1,19 @@
+{ pkgs, ... }:
+
+{
+  system.stateVersion = "23.05";
+
+  networking = {
+    hosts."127.0.0.2" = [ "a" ];
```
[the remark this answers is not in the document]
kurnevsky: Done.

kurnevsky:
It seems formatting is not correct. How do I format it properly?

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Release 23.11 (unreleased)
 
+### New Options
+
+* New options `networking.hosts`, `networking.hostFiles` and
+  `networking.extraHosts` for `/ets/hosts` configuration.
```
t184256, line 8: `/ets` -> `/etc`

t184256 (commented):
I found one typo, but otherwise it looks good to me.

Please fix that one, squash into a commit named like `modules/environment/networking: implement /etc/hosts options` and let's merge.

CHANGELOG.md
kurnevsky: Oops, fixed :)

PR #285, opened 2023-08-12
build: use str instead of string

kurnevsky:
String was deprecated in nixpkgs.

It fails to build with unstable nixpkgs now: https://github.com/NixOS/nixpkgs/pull/247848


t184256:
What about all the other occurrences, should they be replaced as well?

kurnevsky:
That's the only one in this repo :)

t184256:
Oh.

Please also bump the year in the copyright line.

t184256 (approved):

PR #308, opened 2023-11-30
Update README.md

573:
Use nix-community repository for home-manager


t184256 (approved):
Thank you for updating these links.

Issue #221, opened 2022-11-06
Cannot run 'am'


t184256:
I'd like to port https://github.com/DDoSolitary/OkcAgent, but I'm pretty stuck on many fronts.

One is, I cannot run `am` even from the failsafe shell, let alone from with the regular prooted environment:

```
cmd: Failure calling service activity: Failed transaction (2147483646).
```

If resolved, it might open a way to something we've lacked for a long time: tighter integration with Android system.

Somewhat related: https://github.com/t184256/nix-on-droid/issues/3

t184256:
@ettom [shows how do the same from nix-on-droid](https://github.com/t184256/nix-on-droid/issues/185#issuecomment-1659294700):

```
$ /android/system/bin/linker64 /android/system/bin/cmd activity
cmd: Failure calling service activity: Failed transaction (2147483646)
```

t184256:
Note to self: also consider replicating what termux does with their am.apk. With `-b /etc/system/public.libraries.txt`, termux's am.apk gives me:

```
CLASSPATH=from-termux-am.apk LD_PRELOAD=/system/lib64/libnativeloader_lazy.so /android/system/bin/linker64 /android/system/bin/app_process / com.termux.termuxam.Am broadcast -n org.ddsolitary.okcagent/.GpgProxyReceiver
...
Permission Denial: getIntentSender() from pid=..., uid=... (need uid=...) is not allowed to send as package com.termux
```

Could we be close?

t184256:
We now have termux-am available with `android-integration.am.enable = true;`.

PR #311, opened 2023-12-02
Implement termux colorschemes

socksthefennec:
This implements basic support for termux's colorschemes (`~/.termux/colors.properties`) through the new option `terminal.colors`. It takes an attribute set of strings and naively converts that to the key-value format termux expects, e.g. 
```nix
terminal.colors = rec {
  background = color0;
  foreground = color15;
  cursor = color15;

  color0 = "#000000";
  # color1-14 here :)
  color15 = "#FFFFFF";
};
```
would become 
```
background=#000000
color0=#000000
color15=#FFFFFF
cursor=#FFFFFF
foreground=#FFFFFF
```
I wasn't sure whether to use `attrsOf` or `lazyAttrsOf`, so I chose the latter as a guess.
I've mostly just copied over the implementation from what fonts do, I hope that's correct.

There are a couple issues with this as it stands:
1. Colors aren't strictly checked, you can input any string.
2. Attributes aren't checked, you can input any attribute (whereas I believe termux only ever uses background, foreground, cursor and color0-15).
3. There isn't any option to use a preexisting file.
4. The changes don't immediately apply and require opening a new session, but as I understand it that's blocked by #221.
5. Since both font and colors share most of the same code, maybe it could be pulled out into a separate function/functions?
6. My documentation here could probably be better.

It works enough for my usecase, but there's certainly a lot that could be improved.

### DIFF
--- a/modules/terminal.nix
+++ b/modules/terminal.nix
@@ -1,36 +1,47 @@
-# Copyright (c) 2019-2022, see AUTHORS. Licensed under MIT License, see LICENSE.
-
-{ config, lib, ... }:
-
-with lib;
-
-let
+# Copyright (c) 2019-2024, see AUTHORS. Licensed under MIT License, see LICENSE.
+{ config
+, lib
+, pkgs
+, ...
+}:
+with lib; let
   cfg = config.terminal;
 in
-
 {
-
   ###### interface
 
   options = {
-
-    terminal.font = mkOption {
-      default = null;
-      type = types.nullOr types.path;
-      example = lib.literalExpression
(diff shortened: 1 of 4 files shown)


t184256:
Hello and sorry for not getting to this PR for so long.

I took the liberty of rebasing, bumping copyright year, squashing, limiting allowed attrnames and adding some tests. Please take a look.

t184256:
One more thing we'll need is a changelog entry.

Gerschtli (approved):

rahimullahbawar20-crypto:
Ffff

rahimullahbawar20-crypto (commented):
Fddd

PR #323, opened 2024-02-04
add a few more useful commands for easier onboarding

phanirithvij:
When I was setting up ssh access I had to use these commands, I think they are useful as suggestions.

### DIFF
--- a/modules/environment/login/nix-on-droid.nix.default
+++ b/modules/environment/login/nix-on-droid.nix.default
@@ -7,6 +7,8 @@
     vim # or some other editor, e.g. nano or neovim
 
     # Some common stuff that people expect to have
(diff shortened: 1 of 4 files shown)


t184256 (changes requested):
While I tend to agree about procps and killall, openssh, which and file are a step above, IMO. and I don't want to blow this list up too much. I propose adding just procps and killall.

Mind that there's also a fourth copy of this list in `modules/environment/login/nix-on-droid.nix.default`.

phanirithvij:
I made the changes.

phanirithvij:
"changes."
Looks pretty funny in our above comments.

Sorry for this spam comment I couldn't help it 😔.

PR #329, opened 2024-02-15
[WIP] nix.nix: split, align and reuse macros from NixOS

ShamrockLee:
This PR demonstrates the possibility to reuse/inherit NixOS macros.

The original Nix-related options and config in the NixOS modules are mostly inside `nixos/modules/services/system/nix-daemon.nix`, making it hard to reuse without also getting those privilege-required part of implementation. The good news is that the configuration-related part has been split out and become `nixos/modules/config/nix.nix`, `nixos/modules/config/nix-flakes.nix`, `nixos/modules/config/nix-channel.nix` and `nixos/modules/config/nix-remote-build.nix`, making it much easier to reuse.

This PR first align the behavior of `nix.nix` with the NixOS upstream, and then split out `nix-flakes.nix` and `nix-channel.nix`. Finally, it imports the upstream `nix.nix` and `nix-flakes.nix` to replace the local interface and implementation.


modules/environment/nix.nix
```diff
@@ -27,6 +27,15 @@ in
   options = {
 
     nix = {
+      enable = mkOption {
```
t184256, line 35: I don't want to support this option =)

modules/environment/nix.nix
```diff
@@ -50,64 +50,25 @@ in
...
-        trustedPublicKeys = [
-          "cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY="
-          "nix-on-droid.cachix.org-1:56snoMJTXmDRC1Ei24CmKoUqvHJ9XCp+nidK7qkMQrU="
-        ];
-      };
-    }
-  ];
+  config = {
+    nix = {
+      enable = true;
```
t184256, line 62: so, you're introducing .enable, but not using it?

modules/environment/nix.nix
ShamrockLee: It's just there to make implementation from NixOS happy. There is also `{ config.nix.enable = true; }` below to prevent users to turn it off.

Maybe I should remove the `makeIf cfc.enable` stuff inside this `nix.nix`.

modules/environment/nix.nix
ShamrockLee: Yes. I just updated this PR to emphasize this.

ShamrockLee:
`fakedroid` now builds and runs.

The check failure might indicate the need to update the test interface (to supply with the `nixpkgs` argument, etc.). Could you give me some hint where the test cases resides and which file is likely to be responsible for injecting the arguments?

PR #353, opened 2024-05-06
login: Do not dereference proot binds `/bin` and `/etc`

azuwis:
`/bin` and `/etc` are symlinks to `/system/bin` and `/system/etc`, by default proot will dereference and override them with new contents.

After this change, `/system/bin` and `/system/etc` are kept untouched, so commands like `/system/bin/ping` can be run directly.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,3 +2,4 @@ Alexander Sosedkin <monk@unboiled.info>
 Tobias Happ <tobias.happ@gmx.de>
 Bruno Bigras <bigras.bruno@gmail.com>
 Evgeny Kurnevsky <kurnevsky@gmail.com>
+Zhong Jianxin <azuwis@gmail.com>
--- a/modules/environment/login/login.nix
+++ b/modules/environment/login/login.nix
@@ -48,8 +48,8 @@ writeScript "login" ''
 
   exec ${installationDir}/bin/proot-static \
     -b ${installationDir}/nix:/nix \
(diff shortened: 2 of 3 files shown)


t184256:
Nice, this might've been the stuff I've been fighting about a year ago, judging by the age of the [bin-less branch](https://github.com/nix-community/nix-on-droid/compare/master...bin-less)

t184256:
(also, consider adding yourself to AUTHORS)

t184256:
No, actually; with no email notifications from that update of yours, I just forgot about it. Rebased, waiting for the CI...

t184256:
Hm, now tests that check for the absence of some commands fail. If that's because they now find the system ones, why would /system/bin be in `$PATH`?

t184256:
Sorry, the mistake is probably on CI side, as test_channels_uiautomator passes (when tests run properly by typing the command into the app), it's test_channels_shell fails (when tests are run in a hackish way to get better output). I'll see what happens to `$PATH` there...

t184256:
Yeah, a CI problem indeed. Added a commit that sets empty PATH, because just unsetting it isn't enough (something's too smart and sets defaults back), and also unsetting everything messes up more stuff. Let's see whether that solves the issue...

t184256 (approved):

t184256:
Done, thank you for that great fix, and sorry for forgetting about this one!

PR #379, opened 2024-06-13
Emulator-powered testing

t184256:
[x] add building x86_64 proot and bootstrap zipball
[x] figure out flake schema for packages (flat, `prootTermux-aarch64`)
[x] build an x86_64 apk to run tests against, pull it from somewhere (hope we'd just use one from F-Droid next release)
[x] tear fakedroid-based CI out and use an emulator instead (example: https://github.com/t184256/droidctl/commit/b45cbe8b297330d183ad8a05dc2bb343eac0f9e7)
[x] write a droidctl script to bootstrap from a locally built zipball, used as a base for further tests
[x] write a droidctl script to run on-device-tests inside the emulator
[x] write a droidctl script to type several commands into the emulator and check the output (stretch goal)
[ ] test different API levels (stretch goal)
[ ] test aarch64 as well (stretch goal)
[ ] push to cachix from within aarch64 emulators (stretch goal)


modules/build/config.nix
```diff
@@ -12,8 +12,8 @@ with lib;
 
     build = {
       arch = mkOption {
-        type = types.enum [ "aarch64" ];
-        default = "aarch64";
+        type = types.enum [ "aarch64" "x86_64" ];
+        default = strings.removeSuffix "-linux" builtins.currentSystem;
```
Gerschtli, line 16: We should not add impure default values. We can ensure backwards compatibility with this for stateVersion <= 24.05 but should get rid of this impurity in 24.11.

Or we just use nixpkgs.system instead of this property everywhere. Not sure why we are using this arch option instead of just the nixpkgs.system

pkgs/nix-directory.nix
```diff
@@ -25,7 +25,14 @@ stdenv.mkDerivation {
...
     url = "https://nixos.org/releases/nix/nix-2.20.5/nix-2.20.5-${config.build.arch}-linux.tar.xz";
-    sha256 = "sha256:168wjfj3xsc8hq1y6cq59iipjp1g9hmj4n5wdn9c47ad9gbc9cvh";
+    sha256 =
+      let
+        archShas = {
+          aarch64 = "sha256:168wjfj3xsc8hq1y6cq59iipjp1g9hmj4n5wdn9c47ad9gbc9cvh";
+          x86_64 = "sha256:0dax9n562ldj53ap6lz0cwwsfx4d8j1267g9s6lg3zs237yyzw61";
+        };
+      in
+      "${archShas.${config.build.arch}}";
```
Gerschtli, line 35: nit: no need for adding the outer quotes
```
archShas.${config.build.arch}
```

tests/on-device/lib.bash
```diff
@@ -71,3 +71,9 @@ _diff() {
   storePath="$(nix-build "<nixpkgs>" --no-out-link --attr diffutils)"
   "${storePath}/bin/diff" "$@"
 }
+
+detect_system() {
+  nix --experimental-features nix-command \
+      eval --impure --expr 'builtins.currentSystem' \
+    | sed 's|"||g'
```
Gerschtli, line 78: Just use --raw instead of sed :)

flake.nix
```diff
@@ -102,17 +104,26 @@
             (import ./modules {
               inherit extraSpecialArgs home-manager-path pkgs;
               config.imports = modules;
+              config.build.arch =
+                nixpkgs.lib.strings.removeSuffix "-linux" system;
```
Gerschtli, line 108: In the throwIf block above, it is mentioned that system is deprecated and will be removed soon (tm). Maybe don't set a default value for pkgs, then you can just use pkgs.system

Gerschtli (commented):
That is some crazy shit, you've built! Running nix-on-droid on an emulator in GitHub Actions, awesome! Only found some minor things, overall looks very good, thank you :)

tests/on-device/lib.bash
t184256: haha, yeah =)

pkgs/nix-directory.nix
t184256: addressed

flake.nix
t184256: > Maybe don't set a default value for pkgs, then you can just use pkgs.system

OK, I'm confused on how would that work.

nixpkgs-bootstrap aside, there are two more avenues we use and suggest to pass nixpkgs when using flakes:

1. through .follows (see minimal template), and for this one we can't use pkgs.system, since then we just don't know it,
2. through .pkgs (see advanced template), but I don't understand how would it determine one either.

Gerschtli: I was proposing to make `pkgs` (instantiated nixpkgs instance) a mandatory argument instead of `system` to be consistent with home-manager. I am also fine `system` as argument but then it should not trigger a warning when used.

t184256: I think I understand now (45a3da6). Went the "pkgs is mandatory, system errors out" route.

I can make it work, but every time I need to pick between how do I do it, I'm about as good as a coin toss.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
   `pkgs = import nixpkgs { system = "aarch64-linux"; };`
   as an argument to `nixOnDroidConfiguration`.
   If in doubt, refer to the `templates`.
+* `config.arch` option has been dropped. Consider using `pkgs.system` instead.
```
Gerschtli, line 14: Not sure if we need to mention that as this option was marked with `internal` anyways

pkgs/default.nix
```diff
@@ -2,17 +2,18 @@
...
-, arch ? "aarch64"
+, _nativeSystem ? null
 , nixOnDroidChannelURL ? null
 , nixpkgsChannelURL ? null
 , nixOnDroidFlakeURL ? null
 }:
 
 let
-  nixDirectory = callPackage ./nix-directory.nix { };
+  nativeSystem = if _nativeSystem == null then system else _nativeSystem;
```
Gerschtli, line 12: What's the difference between `system`. `_nativeSystem`, `targetSystem` and `pkgs.system`? Could you maybe add some comments in there to explain the intended use cases?

flake.nix
Gerschtli: Yeah true, both ways are fine. I agree on the coin toss regarding which direction to go :D

pkgs/default.nix
t184256: (sorry for the delay) Scattered some comments around.

Gerschtli (approved):

t184256:
Squashed, thank you for the review!

PR #388, opened 2024-07-14
flake.nix: expose `nix-on-droid` package in overlay

pinage404:


t184256:
what's the point of that? what usage does it help with?

pinage404:
When using the nix-on-droid's overlay like in the advanced template

https://github.com/nix-community/nix-on-droid/blob/8bcadcef69dcb5ca177bfb6ea3dc6b092cda2b06/templates/advanced/flake.nix#L43

It allows to install the package with a normal `pkgs.nix-on-droid` instead of `inputs.nix-on-droid.packages."${system}".nix-on-droid`

t184256:
but it is installed by default... I don't get it

pinage404:
it can be usefull if we want to use `nix-on-droid` on another platform

i'm trying to write a deployment script, i use it

t184256:
1. why use that imperative tool from another platform; it's meant to be run on the target device
2. in this case, accessing it from an overlay only makes it harder, doesn't it?

pinage404:
1. my computer is more powerful than my phone, i would like to build on my computer (and check if all is good) instead of pushing, on my phone waiting the long evalution, then see that it fails after a long delay
2. i prefer to add the overlay in one place than writing `inputs.nix-on-droid.packages."${system}".nix-on-droid` everywhere i need `nix-on-droid`

t184256:
sorry, good, but where do you need it at all? how many places is that, what are they?

pinage404:
i haven't finished my script (the feedback loop is super long) for the moment, i need it in 2 places

If you believe, that is not a good thing, i can close this PR

t184256:
I don't know, I need to weigh pros and cons of exposing it this way. Once I do that, it'd be hard to remove it back + I'm not sure what's the behaviour of overriding it in an overlay is etc. etc. Not something I can decide on quickly.

shelvacu:
> my computer is more powerful than my phone, i would like to build on my computer (and check if all is good) 

FYI @pinage404 if you use flakes you can build on your computer with `nix build .#nixOnDroidConfigurations.default.activationPackage --impure`

pinage404:
the current WIP (that doesn't work) https://gitlab.com/pinage404/dotfiles/-/blob/c32e3c71a5662228c30faf4e528354efa8d02843/packages/deploy-fairphone_4/default.nix

why `--impure` is needed ?

t184256:
On July 16, 2024 12:10:53 AM GMT+03:00, pinage404 ***@***.***> wrote:
>the current WIP (that doesn't work) https://gitlab.com/pinage404/dotfiles/-/blob/c32e3c71a5662228c30faf4e528354efa8d02843/packages/deploy-fairphone_4/default.nix
>
>why `--impure` is needed ?
>

Because cross-compiled proot is injected by hash.

shelvacu (changes requested):
I don't think this is the right way of doing this.

If you just want to build your nix-on-droid config somewhere else, run `nix build --impure .#nixOnDroidConfigurations.default.activationPackage`.


If you really want this package installed somewhere other than nix-on-droid, you should do something like:

```nix
# flake.nix
{
  # ...
  inputs.nix-on-droid.url = "github:nix-community/nix-on-droid";
  outputs = { ... }@inputs: {
    nixosConfigurations.yoursystem = nixpkgs.lib.nixosSystem {
      modules = [ {
        environment.systemPackages = [ inputs.nix-on-droid.packages."${system}".nix-on-droid ];
      } ];
    };
  }
}
```

I don't see how an overlay offers any benefits over this approach.

`nix-on-droid` is only meant to run in the nix-on-droid "distro" and may be broken elsewhere. Offering an overlay is just encouraging users to stumble onto such broken behaviour

pinage404:
I understand

Should we close this ?

shelvacu:
yes

PR #389, opened 2024-07-15
modules/environment/session-init: add shellInit option

shelvacu:


shelvacu:
this doesn't actually work, my bad

PR #393, opened 2024-07-26
Grab proot from bootstrap zip rather than including its nix path

shelvacu:
This means that the cachix substituter (or already having the package in your nix store somehow) is no longer required to build.

This required reworking the deploy script. As a bonus you can now omit the second argument and it will tell you what it would have copied instead of copying anything.

This is fixes one source of impurity, but for now flake builds will still require the --impure flag


modules/environment/login/proot-attrs/aarch64.nix
```diff
@@ -0,0 +1,5 @@
+# WARNING: This file is autogenerated by the deploy script. Any changes will be overridden
+{
+  url = "file:///data/local/tmp/n-o-d/bootstrap-aarch64.zip";
```
t184256, line 3: I presume, the committed version will have my bootstrap url here instead?

Isn't there a dependency loop then, with the bootstrap tarball containing the hash of itself?

shelvacu: "will have my bootstrap url here instead" yes

"isn't there a dependency loop"
No, two things:
- the hash is only a hash of the proot binary
- during bootstrap building prootStatic is overridden, so this is never evaluated

shelvacu:
updated to fix formatting

scripts/deploy.sh
```diff
@@ -51,27 +55,45 @@ if [[ "$PUBLIC_URL" =~ ^file:///(.*)/archive.tar.gz ]]; then
...
     log "building $arch proot..."
     proot="$(nix build --no-link --print-out-paths ".#prootTermux-${arch}")"
+    proot_hash="$(nix-hash --type sha256 --sri "$proot"/bin/proot-static)"
+    attrs_file="modules/environment/login/proot-attrs/$arch.nix"
+    new_attrs_file="$(mktemp)"
+
+    cat >$new_attrs_file <<EOF
+# WARNING: This file is autogenerated by the deploy script. Any changes will be overridden
+{
+  url = "$NIX_ON_DROID_BASE_URL/bootstrap-$arch.zip";
```
t184256, line 75: this lacks a branch (github:owner/repo/branch case, see above)

t184256: No, wait, it generally isn't inferrable. The interface will have to be extended to pass the public url of the zip tarball as well =(

scripts/deploy.sh
```diff
@@ -51,27 +55,45 @@ if [[ "$PUBLIC_URL" =~ ^file:///(.*)/archive.tar.gz ]]; then
...
-        exit 1
+        log "updating contents of $attrs_file"
+        echo "<<<<<<"
+        cat $attrs_file
+        echo "======"
+        mv $new_attrs_file $attrs_file
+        cat $attrs_file
+        echo ">>>>>>"
+        log "adding $attrs_file to git index"
+        git add $attrs_file
```
t184256, line 96: I don't think we should.

t184256:
> This is fixes one source of impurity, but for now flake builds will still require the --impure flag

And what's the other impurity source?

scripts/deploy.sh
shelvacu: I think we have to, otherwise it's not included in the flake. I agree it's less than ideal but I don't see an obvious alternative.

shelvacu:
> > This is fixes one source of impurity, but for now flake builds will still require the --impure flag
> 
> And what's the other impurity source?

- Reading from env vars
- uid/gid
There may be others but those are the two I'm aware of

scripts/deploy.sh
t184256: The use case I want covered is, basically, `someuser` cloning nix-on-droid, creating a branch, changing it and then wanting to test it without adb and with a HTTP server. while we know where to pull the github source tarball from if it's on github (and non-github users assumed to know what they're doing), we don't know what URL will they place the bootstrap tarball at. previously that didn't matter to us, as long as they enter it correctly into the app; now it begins to matter

t184256:
I've made one reading pass through it, I am willing to explore that direction in general, but I'm afraid I'll make a million little opinionated edits over the course of the next ~10 days.

shelvacu:
Opinionated edits aren't necesarily bad, I believe I checked the box so that you can directly push changes to the branch if you want to.

t184256:
So far the only global consideration I've found for my workflows is that I can't update proot in between releases, lest we'll have master commits using a version of proot with no stable-path zipball to back them.

shelvacu:
> So far the only global consideration I've found for my workflows is that I can't update proot in between releases, lest we'll have master commits using a version of proot with no stable-path zipball to back them.

I don't know what you want from me on this one. If you do it anyways it's no worse off than before (sometimes people will need to have the cachix substituter), or you can follow a nightly-style filename format (eg `/bootstrap-unstable-2024-09-01/bootstrap-aarch64.zip`, would also take up more space ofc), or you switch to a hash-only (no url) fixed-output derivation.

scripts/deploy.py
```diff
@@ -0,0 +1,175 @@
...
+    # for CI and local testing
+    if (m := re.search("^file:///(.*)/archive.tar.gz$", public_url)) is not None:
+        flake_url = f"/{m[1]}/unpacked"
+    else:
+        flake_url = public_url
+    base_url = re.sub("/[^/]*$", "", public_url)
+    if bootstrap_url is None:
+        bootstrap_url = base_url
+
+    log(f"channel_url = {channel_url}")
```
t184256, line 106: sorry, were these four calculated and then never used?

shelvacu: yes, oops. will fix when i get a chance

shelvacu: fixed

PR #412, opened 2024-09-10
tests/emulator: add license, lint with ruff

t184256:
Adds ruff-based linting for python files.

Related to @shelvacu's https://github.com/nix-community/nix-on-droid/pull/393,
where I do appreciate the idea of linting, but I'd like to have it decoupled from script-building time.


Gerschtli (approved):
The new formatting LGTM, though I am not that familiar with ruff or its alternatives. But basic linting and formatting is always a good thing :)

.github/workflows/lints.yml
```diff
@@ -17,3 +17,19 @@ jobs:
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install nix
+        uses: cachix/install-nix-action@v25
+
+      - name: Run ruff linter
+        run: nix run 'nixpkgs#ruff' -- check
```
shelvacu, line 32: in the interest of running the version of ruff we expect to run, shouldn't this point to something in `flake.nix`? Unfortunately I don't think `nix run .#nixpkgs.legacyPackages.x86_64.ruff` works but we could expose ruff in the flake and run that.

shelvacu (approved):
minor nitpick but looks good

t184256:
hm. maybe it'd be more logical to pack it into a flake check than to expose ruff from it? I don't have experience with flake checks though.

Gerschtli:
in the end, a flake check is just a derivation that either ends with `touch $out` in case of success or with `exit 1` in case of error.

PR #442, opened 2025-02-16
Install on vanilla termux

izelnakri:
This PR is incomplete, instead we should create another bootstrap zip for termux specifically while respecting existing nix-on-droid-app code contract. 

I created a PR here to show you @t184256 the only working bootstrap zip for termux, after 2+ hours of manual cross-compilation from my machine to scp it and set it up on the android device:

```sh
[pc:~/nix-on-droid]$ NIX_ON_DROID_CHANNEL_URL=https://github.com/KoviRobi/nix-on-droid/archive/refs/heads/install-on-vanilla-termux.tar.gz \
                     NIX_ON_DROID_FLAKE_URL=github:KoviRobi/nix-on-droid/install-on-vanilla-termux \
                     nix build ".#bootstrapZip-aarch64" --impure
[android:~]$ scp pc:nix-on-droid/result/bootstrap-aarch64.tar.gz .
[android:~]$ mkdir ../nix/
[android:~]$ tar xf ~/bootstrap-aarch64.tar.gz -C ../nix/
[android:~]$ chmod +w -R ../nix/
[android:~]$ ../nix/bin/login
```

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -7,11 +7,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1709445365,
-        "narHash": "sha256-DVv6nd9FQBbMWbOmhq0KVqmlc3y3FMSYl49UXmMcO+0=",
+        "lastModified": 1732482255,
+        "narHash": "sha256-GUffLwzawz5WRVfWaWCg78n/HrBJrOG7QadFY6rtV8A=",
         "owner": "nix-community",
         "repo": "home-manager",
-        "rev": "4de84265d7ec7634a69ba75028696d74de9a44a7",
+        "rev": "a9953635d7f34e7358d5189751110f87e3ac17da",
         "type": "github"
       },
       "original": {
@@ -31,11 +31,11 @@
         "nmt": "nmt"
       },
       "locked": {
-        "lastModified": 1705252799,
-        "narHash": "sha256-HgSTREh7VoXjGgNDwKQUYcYo13rPkltW7IitHrTPA5c=",
+        "lastModified": 1729445390,
+        "narHash": "sha256-TxJ7RZLlBkKWZos1ai3eWIH0fBq1G6SVE+q3dW+0qRU=",
         "owner": "Gerschtli",
         "repo": "nix-formatter-pack",
-        "rev": "2de39dedd79aab14c01b9e2934842051a160ffa5",
+        "rev": "9f4bcf647cad2edafda7e1143071e0daf37cbc41",
         "type": "github"
       },
       "original": {
@@ -46,31 +46,15 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1708172716,
-        "narHash": "sha256-3M94oln0b61m3dUmLyECCA9hYAHXZEszM4saE3CmQO4=",
+        "lastModified": 1732665169,
+        "narHash": "sha256-BWzlDJpfcE7DgxtAkeXHnU69Iz6cU64yCgE6b+hqds8=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "5d874ac46894c896119bce68e758e9e80bdb28f1",
+        "rev": "69661586d1c9d85e859887d95285f6fc92521657",
         "type": "github"
       },
       "original": {
         "owner": "NixOS",
-        "repo": "nixpkgs",
-        "type": "github"
-      }
-    },
-    "nixpkgs-docs": {
-      "locked": {
-        "lastModified": 1705957679,
-        "narHash": "sha256-Q8LJaVZGJ9wo33wBafvZSzapYsjOaNjP/pOnSiKVGHY=",
-        "owner": "NixOS",
-        "repo": "nixpkgs",
-        "rev": "9a333eaa80901efe01df07eade2c16d183761fa3",
-        "type": "github"
-      },
-      "original": {
-        "owner": "NixOS",
-        "ref": "release-23.05",
         "repo": "nixpkgs",
         "type": "github"
       }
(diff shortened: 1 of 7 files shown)


t184256:
Quick question: what's the benefit of doing all of that over [nix-in-termux](https://github.com/t184256/nix-in-termux)?

izelnakri:
Hi @t184256 , I explained the reasons here: https://github.com/nix-community/nix-on-droid/issues/257#issuecomment-2661580637

Today it is VERY difficult to install nix on termux and it is manual/not automated, so we need to have this bootstrap zip hosted on the web server with the current release pipeline. Then one can even package the installation up as a Termux package: `$ pkg install nix-on-droid`

t184256:
>Hi @t184256 , I explained the reasons here: https://github.com/nix-community/nix-on-droid/issues/257#issuecomment-2661580637

No, this is about "why use nix-on-droid and Nix on Android in general".

>Today it is VERY difficult to install nix on termux and it is manual/not automated,

And I've linked to a project automating that with much fewer code than nix-on-droid.

> so we need to have this bootstrap zip hosted on the web server with the current release pipeline. Then one can even package the installation up as a Termux package: `$ pkg install nix-on-droid`

Sorry, I see 0 connection between bootstrap zips and termux debs. Why have the first if you can build the second?

izelnakri:
HI again @t184256 ,

I'm aware half of the development is proper communication & I try. Probably the best methodology/result is actually providing the changes that will make it more clear what specific changes I want.

My first 4 points explain the reasons why nix on termux instead of `nix-on-droid` a better idea:

```
a - Nix-on-droid app uses a fork of Termux which backtracks new features, changes & fixes to current Termux, if it ever gets these changes.
b - Nix-on-droid app doesn't support Termux addons which are essential if you want to run scripts/binaries in your shell environment on device boot (Termux:Boot) or service management(like previously discussed supervisord, runit or systemd: Termux:Services)
c - Nix-on-droid app doesn't support Termux RUN_COMMAND Intent interface which is required if you want to run a binary installed on users system from users another android app via Android Intent RPC framework.
d - Nix-on-droid repo targets com.termux.nix not com.termux, installing nix-on-droid on Termux is NOT trivial currently, whereas it could. Currently, manually compiling fork with very specific changes & build instructions for boostrap zip took me several hours to get a successful result after many different failed trials.
```

t184256:
Yeah, something seems broken with our communication here, but that's fine =) Let me try again:

You seem to be answering "why would one want to forego nix-on-droid Termux fork app and use Termux with Nix installed inside it".
You provide great, compelling reasons detailing how the regular Termux app user experience is better than my fork's. I respect that.
In fact, [my first attempt to run Nix on my very first Android device](https://monk.unboiled.info/pub/talks/2019/nixcon-nix-on-droid.video.mp4), a cute [~50 SLOC script called](https://github.com/t184256/nix-in-termux/blob/master/nix-in-termux) employed exactly the approach you're detailing here. It took Nix and installed it under Termux, using proot to keep Nix store at /nix/store and avoid expensive recompilation. Could've also used root, but I disgress.

Later on I've decided I don't use Termux other than for installing Nix and started a different project, nix-on-droid, that skips right past the Termux part and bootstraps from my entirely Nix-built zipball. Would be cool if Termux app allowed me to specify an url and do a couple more things differently wrt discovering the entry point. It didn't, so I forked it under nix-on-droid-app.

Six years later, you send a pull request to nix-on-droid, a project that's now easily 10x the size of nix-in-termux, that (to the best of my understanding, sorry if I understood something wrong here) does a very roundabout thing: uses this monstrosity, effectively, as an archive with Nix inside it, so that you can unpack it into your Termux environment and enjoy using Nix from there.

As I said, a valid goal in itself, except there's one thing I absolutely do not understand. There already exists an archive with Nix and all the required dependencies inside it that one can unpack and use, an official Nix release tarball. A tarball that nix-on-droid uses as a starting point for its zipball. A tarball that nix-in-termux downloads and unpacks. Half of the script is downloading it and unpacking it. Half of the script is prooting into the resulting environment. That's literally all to it. You can just do the same, maybe fork my nix-in-termux repo, update the urls/hashes and maintain it in its 50 SLOC glory. Unless I'm missing something?

Instead you file a pull request to nix-on-droid, a much larger project with an established userbase, to make it suit your needs, one that'd render their installs broken. With no tests, no upgrade paths, no nothing.

Hope that clarifies my confusion about your request: why this PR to nix-on-droid instead of reviving nix-in-termux?

Or doing that thing you've proposed with packaging nix as part of Termux. That sounds neat.

izelnakri:
@t184256 Thanks for the very good explanation! This PR is not finished as you correctly pointed out it doesnt even include any tests or finished solution also I certainly appreciate & respect your solution when it comes installing nix on android devices.

>In fact, [my first attempt to run Nix on my very first 64-bit Android device](https://monk.unboiled.info/pub/talks/2019/nixcon-nix-on-droid.video.mp4), a cute [~50 SLOC script called nix-in-termux](https://github.com/t184256/nix-in-termux/blob/master/nix-in-termux) employed exactly the approach you're detailing here. It took Nix and installed it under Termux, using proot to keep Nix store at /nix/store and avoid expensive recompilation. Could've also used root, but I digress.

I tried the 50LOC script and it didnt work for me **yesterday**, dont remember the specific error unfortunately but I **really** tried building this with that script, with my fork with changes etc. **Only** these changes worked.

> You can just do the same, maybe fork my nix-in-termux repo, update the urls/hashes and maintain it in its 50 SLOC glory. Unless I'm missing something?

I did precisely that, it didn't work. So I had to manually compile the bootstrap zip myself and wait **hours** of compilation and unpack it to my android device. Thus I suggest streamlining this process by adding a termux zip/tar to the release/web server pipeline.

> "why this PR to nix-on-droid instead of reviving nix-in-termux"?

I also suggest removal of the nix-on-droid android app/simplifying/removing a lot of code here. I come from Ruby background where we had tons of ways to solve a problem, later in my life I realized its is MUCH better solution to remove code from a system whenever possible and support the best solution more and potential remove the curse of choice for others. Thus I also suggest to endorse "nix-in-termux" approach more, but thats of course up to you.

I know PR is still incomplete, let me know if you'd like me to have a look for the right needed changes. I'll probably get back to this PR but it might take few days/weeks.

Thanks again for this awesome solution! I just think it can be better :pray:

PR #443, opened 2025-02-18
modules/user: allow users to select their username

### LINKED ISSUE #320: Change userName?
I found the option for setting the username of the account used.

https://github.com/nix-community/nix-on-droid/blob/039379abeee67144d4094d80bbdaf183fb2eabe5/modules/user.nix#L57

But it is set to readonly.  Is there an important reason that's not configurable, or is there some other avenue to set it?

osbm:
Fixes #320 

Maybe we can do this for these options too, but i am not sure what could be gained by being able to change them.
https://github.com/nix-community/nix-on-droid/blob/5d88ff2519e4952f8d22472b52c531bb5f1635fc/modules/user.nix#L93-L95

### DIFF
--- a/modules/user.nix
+++ b/modules/user.nix
@@ -28,7 +28,7 @@ in
     user = {
       group = mkOption {
         type = types.str;
-        readOnly = true;
+        default = "nix-on-droid";
         description = "Group name.";
       };
 
@@ -56,7 +56,7 @@ in
 
       userName = mkOption {
         type = types.str;
-        readOnly = true;
+        default = "nix-on-droid";
         description = "User name.";
       };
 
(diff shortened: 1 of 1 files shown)


osbm:
What are your thoughts @t184256 @Gerschtli ?

Gerschtli:
I think we can also make the group name configurable for consistency. The home dir I wouldn't change because migrating that is not trivial.

I can't see any problem by changing the names but I don't have the time to test it.

osbm:
Shall i also add the group name too?

osbm:
@Gerschtli done.

Gerschtli (approved):

PR #457, opened 2025-05-10
nix-on-droid: passthrough --override-input arg

panchoh:
With this addition, we would simplify the development cycle of stylix modules when using flakes:

`nix-on-droid switch --flake . --override-input stylix git+file:/home/user/path/to/stylix`

since we wouldn't need to to do a `nix flake lock --update-input stylix` for every edit to the local stylix input.

### DIFF
--- a/nix-on-droid/nix-on-droid.sh
+++ b/nix-on-droid/nix-on-droid.sh
@@ -68,6 +68,7 @@ function doHelp() {
     echo "  --keep-going"
     echo "  --max-jobs NUM"
     echo "  --option NAME VALUE"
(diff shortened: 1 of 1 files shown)


Gerschtli (commented):
Please document it in the --help output, too :)

panchoh:
> Please document it in the --help output, too :)

Oh, right!
On it!

panchoh:
> Please document it in the --help output, too :)

Done! (Sorry I didn’t think about it before).

Gerschtli (approved):

PR #468, opened 2025-06-29
Add environment.extraSetup

WiiPlayer2:
This adds a new option `environment.extraSetup` like the one already in NixOS.

I need this for the same reason as described in the option: to generate MIME cache etc. for desktop environments.


t184256:
Is the same achievable with [activation](https://nix-community.github.io/nix-on-droid/nix-on-droid-options.html#opt-build.activation`)?

PR #470, opened 2025-07-11
feat: convert ruff lints to flake checks

plaflamme:
This is based on https://github.com/nix-community/nix-on-droid/pull/412 and is an answer to [this comment](https://github.com/nix-community/nix-on-droid/pull/412#issuecomment-2344202710)

Notes:

1. probably not the most idiomatic nix, but it works (try modifying one of the `.py` files and then running `nix flake check --impure`)
2. had to use `--impure` argument, I'm not sure why the error kicks in (see 505cc019aef7c217c700b61536adcb22d0f7b4d1), but it complains about referencing `builtins.storePath`
3. modified the GH action to simply invoke `nix flake check --impure` which also runs the pre-existing nix linting check
4. did `nix flake update` to get a more recent `ruff` version (0.2.x -> 0.12.x which is pretty substantial)

cc @t184256

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -7,11 +7,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1709445365,
-        "narHash": "sha256-DVv6nd9FQBbMWbOmhq0KVqmlc3y3FMSYl49UXmMcO+0=",
+        "lastModified": 1752256062,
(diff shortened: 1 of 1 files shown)


plaflamme:
@t184256 Any feedback? I'm wondering if this can help get #412 merged which can then unblock #393

t184256:
wonder why it's rebuilding so much...

PR #471, opened 2025-07-12
Pass class="nixOnDroid"; to evalModules

shelvacu:
This is useful for getting better error messages when mixing modules; a module designed for nixos configs sets `_class = "nixos";` and then when you try to use it from a nix-on-droid config it shows an error.

See: https://nixos.org/manual/nixpkgs/unstable/#module-system-lib-evalModules-param-class

### DIFF
--- a/tests/on-device/config-class.bats
+++ b/tests/on-device/config-class.bats
@@ -1,0 +1,25 @@
+# Copyright (c) 2023, see AUTHORS. Licensed under MIT License, see LICENSE.
+
+load lib
+
+@test 'successfully loads a config with _class="nixOnDroid"' {
+  # set up / build / activate the configuration
+  echo '{ config.system.stateVersion = "24.05"; _class = "nixOnDroid"; }' > ~/.config/nixpkgs/nix-on-droid.nix
+  _sed -e "s|<<FLAKE_URL>>|$FLAKE_URL|g" -e "s|<<SYSTEM>>|$(detect_system)|g" \
+    "$ON_DEVICE_TESTS_DIR/config-flake.nix" \
+    > ~/.config/nixpkgs/flake.nix
+
+  nix-on-droid switch --flake ~/.config/nixpkgs#device
+}
+
+@test 'fails to load a config with _class="nixos"' {
+  # set up / build / activate the configuration
+  echo '{ config.system.stateVersion = "24.05"; _class = "nixos"; }' > ~/.config/nixpkgs/nix-on-droid.nix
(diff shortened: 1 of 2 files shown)


t184256:
Could you also add a test?

shelvacu:
I'm not sure what you would want a test for this to look like. Like, invoke nix on a 'bad' config and then grep the output for the right error message? That sounds very fragile.

t184256:
Yeah, something like that.

The closest test I could find just checked for erroring out, but then it did so on a small change compared to a previously applied working config. I guess that's not an option here.

shelvacu:
which test are you looking at?

tests/on-device/config-class.bats
```diff
@@ -0,0 +1,25 @@
...
+}
+
+@test 'fails to load a config with _class="nixos"' {
+  # set up / build / activate the configuration
+  echo '{ config.system.stateVersion = "24.05"; _class = "nixos"; }' > ~/.config/nixpkgs/nix-on-droid.nix
+  _sed -e "s|<<FLAKE_URL>>|$FLAKE_URL|g" -e "s|<<SYSTEM>>|$(detect_system)|g" \
+    "$ON_DEVICE_TESTS_DIR/config-flake.nix" \
+    > ~/.config/nixpkgs/flake.nix
+
+  # check that networking.hosts can't map localhost
```
t184256, line 22: looks like a leftover comment

tests/on-device/config-class.bats
```diff
@@ -0,0 +1,25 @@
...
+  _sed -e "s|<<FLAKE_URL>>|$FLAKE_URL|g" -e "s|<<SYSTEM>>|$(detect_system)|g" \
+    "$ON_DEVICE_TESTS_DIR/config-flake.nix" \
+    > ~/.config/nixpkgs/flake.nix
+
+  nix-on-droid switch --flake ~/.config/nixpkgs#device
+}
+
+@test 'fails to load a config with _class="nixos"' {
+  # set up / build / activate the configuration
+  echo '{ config.system.stateVersion = "24.05"; _class = "nixos"; }' > ~/.config/nixpkgs/nix-on-droid.nix
```
t184256, line 17: I was thinking of less synthetic examples of somehow (accidentally?) trying to use nixos modules, but I'm not sure how cumbersome that'd be to write.

tests/on-device/config-class.bats
shelvacu: fixed

tests/on-device/config-class.bats
t184256: ok then, but then it's not that useful yet, I guess

t184256:
that was no magic, it was a flaky test I re-ran manually =/

t184256 (approved):

PR #482, opened 2025-10-15
nix-on-droid: passthrough --log-format arg

keysmashes:
based on #457.

now you can `nix-on-droid --log-format multiline switch` to see _all_ the things that are currently building, or use `multiline-with-logs` to see the build logs too.


Gerschtli:
nix also has a --log-format option

mio-19:
json is valid for nix. this would be useful with nix output monitor

Yeshey:
would love to use `--log-format` with nix-on-droid

PR #483, opened 2025-10-16
Don't use example.org in tests

keysmashes:
The text at example.org was recently changed, which breaks the emulator tests. There is now a request there to "Avoid use in operations", so using something else in the tests seems like the friendly thing to do.

(checking if the test actually passes in CI before I unmark this as draft...)


keysmashes:
well, `android_integration` now passes, I don't know what's up with `test_channels_shell` :melting_face:

t184256 (approved):

PR #503, opened 2026-02-17
fix: proot-termux LLVM 21 build and runtime fixes

mhuesch:
## Summary

Fixes the proot-termux cross-compilation and runtime crashes under LLVM 21
(nixpkgs 25.11), unblocking the 25.11 upgrade.

Three issues were identified and fixed:

### 1. Cross-prefixed `readelf` (build fix)
The GNUmakefile invokes bare `readelf -s` to generate `loader-info.c`, but
only the cross-prefixed binary exists in the cross toolchain. Without this,
`offset_to_pokedata_workaround` silently defaults to 0.

### 2. Loader ELF `p_align` (EINVAL on execve)
LLVM 21's `lld` sets `file_offset = vaddr` when `-Ttext=0x2000000000` is
used, producing an 8 GB loader binary. Adding `-n` (nmagic) to the linker
flags fixes the file size but sets `p_align=4` in the LOAD program header.
Android kernels require page-aligned LOAD segments and reject `p_align=4`
with `EINVAL` on `execve(2)`.

Fix: link with `-n` for compact layout, then binary-patch `p_align` to
`0x10000` (64 KB) in the loader ELF. This supports 4 KB, 16 KB, and 64 KB
page-size devices.

### 3. Stack pointer alignment in `transfer_load_script` (SIGBUS)
**Root cause of the runtime crash.** After `execve`, proot sets the tracee's
stack pointer to `stack_pointer - buffer_size`. On AArch64, SP must be
16-byte aligned. LLVM 17's generated loader code happened to tolerate
misaligned SP, but LLVM 21 emits SP-relative addressing in the loader's
`_start` prologue, causing an immediate `SIGBUS` (`BUS_ADRALN`) at
`PC=0x2000000000` (the loader entry point).

Fix (`align-sp.patch`): round SP down to 16-byte alignment independently
of the load script data pointer (x0/USERARG_1). SP and x0 must differ —
if both are set to the aligned address, the absolute string pointers
embedded in the load script become offset by the alignment delta.

This is a bug in upstream proot-termux (`src/execve/exit.c`) that should
be reported/upstreamed.

## Testing

Tested on a OnePlus 9 (Snapdragon 888, aarch64, Android 14) with an
existing nix-on-droid installation:
- `proot-static --version` ✓
- `proot -b .../nix:/nix /bin/sh -c 'echo hello'` → prints "hello", exit 0 ✓
- Full nix-on-droid login flow via `.proot-static.new` staging mechanism ✓

## Files changed

- `pkgs/proot-termux/default.nix` — all three workarounds in `preConfigure`
- `pkgs/proot-termux/align-sp.patch` — new, patches `src/execve/exit.c`


mhuesch:
yesterday I was using Claude Code to update my nixpkgs, and I noticed the 25.11 prerelease work.

knowing little about the innards of this repo, but being quite impressed recently with CC's ability to figure stuff out, I decided to turn it loose on the tip of `prerelease-25.11` today and spent a number of hours shepherding it along and copy/pasting commands and output between my old Android phone and my desktop.

CC did basically all the real work here, I was just "managing" and vaguely following along while ferrying commands/output to the test device. I'm quite impressed.

I hope this patch is helpful.

t184256:
> On AArch64, SP must be 16-byte aligned. LLVM 17's generated loader code happened to tolerate misaligned SP, but LLVM 21 emits SP-relative addressing in the loader's _start prologue, causing an immediate SIGBUS (BUS_ADRALN) at
PC=0x2000000000 (the loader entry point).

Holy flying Batman.

Let's see what would the upstream say about it.

petm5:
Could this be the cause of the bus errors when linking against anything other than bionic libc? (https://github.com/nix-community/nix-on-droid/pull/18#issuecomment-538699770)

mhuesch:
maybe this would be predictable to someone with more experience (lol), but I tried to use this PR's branch to bump my nix-on-droid to 25.11 on my Android phone. I was able to get it build and `nix-on-droid switch --flake` to it. but then exiting and restarting the app caused a crash and I had to rollback manually via failsafe.

I assume there's more work needed in the runtime to sync up w/ the 25.11 nixpkgs-level updates.

if there's more (Claude Code and) I can do to help to get this across the line, please let me know, happy to help.

Screenshot_20260228-222949

mhuesch:
@t184256 sorry I only just saw your comment https://github.com/nix-community/nix-on-droid/issues/502#issuecomment-3980483585! funny timing as I did that attempt last night and then aborted/rolled-back this morning. great news, thank you!

PR #505, opened 2026-02-22
Standardize Cross-Compilation and Enable Pure Flakes

petm5:
These changes refactor the build pipeline for `talloc`, `proot-static`, and the bootstrap zipball to use standard nixpkgs cross-compilation infra. By doing so, we remove the need for custom cross-compilation logic and allow system configurations without the `--impure` flag.

**Highlights:**
- Nixpkgs is updated to take advantage of general cross-compilation improvements.
- `proot` is updated to prevent `SIGBUS` crashes on aarch64 when built against musl and recent LLVM (https://github.com/termux/proot/commit/28baec5ac3e0f26c03cf2a4cdc094c9328bd5b20).
- The hard-coded `proot` and `talloc` store paths in `nixOnDroidConfiguration` are replaced with a nixpkgs-native LLVM/musl cross toolchain. The same toolchain is used to build the bootstrap zipball.
- Modified the `talloc` package so it can be built outside of a cross-compilation environment.
- Removed the custom cross-compiler infrastructure and cross-compiled package specifications as they are no longer necessary.
- Nix store initialization is now handled on-device during initial setup in `login-inner`. This lets us avoid duplicating the logic of `nix-store --init`.
- The hard-coded nix release tarball is replaced with a native package from nixpkgs.

It is possible to specify what architecture will be used to cross-compile the static components. The idea is they could be built on an `x86_64-linux` system and stored in a binary cache so aarch64 devices can use them even if they can't be built natively.

The changes were tested and validated on my own device (Samsung Galaxy A15) using a custom binary cache and bootstrap server. I could switch to a home-manager enabled profile without it throwing bus errors.

Further changes may be needed to maintain compatibility with the release scripts and GitHub Actions checks.


t184256:
That's very interesting, should be landed after a release once I finally force myself to make one.

Note to self: I also wonder whether it's posible to hop off musl/LLVM altogether with that SIGBUS fix. IIRC, the symptom that made us abandon glibc/GCC 7 years ago was also manifesting itself as an inexplicable SIGBUS: https://github.com/nix-community/nix-on-droid/issues/6#issuecomment-481316644

petm5:
> I also wonder whether it's posible to hop off musl/LLVM altogether

AFAIK the whole point of using musl is to enable static linking since glibc doesn't support it well. How would a dynamically-linked proot work without hacks like a custom Nix store path?

t184256:
sorry, right, I meant bionic/LLVM -> musl/GNU, the default for nixpkgs static compilation

PR #510, opened 2026-03-09
Fixed-Output Derivation for Bootstrap Tarball

tennox:
Makes the Nix tarball download in bootstrap package build reproducible and deterministic by using Fixed-Output Derivations (FOD) - instead of impure `builtins.fetchurl`.


tennox:
I realized the whole nixDirectory import is very impure and this doesn't help much...
so can be merged but doesn't change much I guess

PR #523, opened 2026-04-12
activation: Get activation package with --print-out-paths

petm5:
Nix is currently called a second time after the activation package build to find its path in the store. This causes the nix-on-droid config to be evaluated twice, adding overhead and slowing down iteration.

Modern Nix outputs build logs to STDERR and capturing the path from STDOUT should be safe. Only the first line is used in case the config outputs multiple paths for some reason.


frankitox:
This is great! My build time halfed 30s -> 15s.

Issue #515, opened 2026-03-23
stdin is not a terminal


MakiseKurisu:
Recently switched my phone to OnePlus Ace 3V, and on this phone I can no longer run programs that need a working terminal. This includes nano and ssh.

A recording showing the difference between termux and nix-on-droid. Both are the latest version from F-Droid.

https://github.com/user-attachments/assets/6111ada7-5de7-405a-8b2a-b24ff7d16809

Gipphe:
I'm experiencing the same issue with other editor like Neovim and Vim where they report that `stdin` and `stdout` are not terminals.

Now, feel free to call me out if this is completely off (especially since `proot` is a bit special), but is `proot` supposed to remain the session leader instead of the running shell?

`ps` gives me this on `nix-on-droid`:

```
bash-5.2$ ps -o pid,pgid,sid,tty,comm
  PID  PGID   SID TT       COMMAND
18006 18006 18006 pts/0    proot-static
18015 18015 18006 pts/0    bash
31562 31562 18006 pts/0    ps
```

And this on regular Termux:

```
u0_a342@localhost ~> ps -o pic,pgid,sid,tty,comm
  PID  PGID   SID TT       COMMAND
12652 12652 12652 pts/1    bash
12708 12708 12652 pts/1    ps
```

Is that to be expected when using `proot`?

t184256:
On April 10, 2026 12:13:40 PM GMT+02:00, Victor Nascimento Bakke ***@***.***> wrote:
>Gipphe left a comment (nix-community/nix-on-droid#515)
> is `proot` supposed to remain the session leader instead of the running shell?

no idea

>`ps` gives me this on `nix-on-droid`:
>
>```
>bash-5.2$ ps -o pid,pgid,sid,tty,comm
>  PID  PGID   SID TT       COMMAND
>18006 18006 18006 pts/0    proot-static
>18015 18015 18006 pts/0    bash
>31562 31562 18006 pts/0    ps
>```

>And this on regular Termux:
>
>```
>~ $ ps -o pic,pgid,sid,tty,comm
>  PID  PGID   SID TT       COMMAND
>12652 12652 12652 pts/1    bash
>12708 12708 12652 pts/1    ps
>```
>
>Is that to be expected when using `proot`?

That's expected when not using proot =) `proot bash` and compare?

Gipphe:
> That's expected

[...]

PR #529, opened 2026-05-05
proot-termux: unstable-2024-05-04 -> unstable-2026-02-20

### LINKED ISSUE #495: Activation failed with the latest nixos-unstable
```

Building activation package...
Executing activation script...
Activating linkBinSh
Activating linkUsrBinEnv
Activating setPriorityHomeManagerPath
Activating installLogin
Activating installLoginInner
Activating installPackages
replacing old 'nix-on-droid-path'
installing 'nix-on-droid-path'
error: getting pseudoterminal attributes: Permission denied
error:
       … while waiting for the build environment for '/nix/store/f8rrg4q0mbvkidwj2ywrsmnbivjw52b5-user-environment.drv' to initialize (failed with exit code 1, previous messages: )

       error: reading a line: Input/output error
```

It was fine with nixos-unstable from 2026-01-24

see nixos-unstable changes between 2026-01-24 and 2026-01-31

https://github.com/NixOS/nixpkgs/compare/88d3861acdd3d2f0e361767018218e51810df8a1...bfc1b8a4574108ceef22f02bafcf6611380c100d

nix 2.31.2 -> 2.31.3

https://github.com/NixOS/nix/compare/2.31.2..2.31.3

A user reported that there is no issue with determine nix

tried `nix.package = pkgs.nixVersions.nix_2_30;`, same error

newAM:
Fixes #495

The proot-termux update contains patches to fix the TCSETS2 syscall used by `tcgetattr`.

- [add workarounds for TCSETS2 and TCSETSF2](https://github.com/termux/proot/commit/228a5f28b078f4e2504de46758ce17948f73f507)
- [replace TCSETSW2 ioctl with TCSETSW](https://github.com/termux/proot/commit/2cb63b836ad7eae0d7f1fd3c941848a84acb5dcc)
- [replace TCGETS2 ioctl argument with TCGETS](https://github.com/termux/proot/commit/6fa36f7b4405dc08efc87139c0d926a3a3f98fd0)

I have tested this on a fresh nix-on-droid install using the F-Droid apk.  Steps to test: 

1. Unpin the `nixpkgs` input for `nix-on-droid` in your `flake.nix` (assume you pinned to workaround #495)
2. Build `nix build .#prootTermux-x86_64` and `nix build .#prootTermux-aarch64`
3. ~~Copy the proot builds to a public nix cache, put your cache URL and key in `modules/environment/nix.nix`~~
4. Build `nix build ".#bootstrapZip-aarch64" --impure` and copy the resulting .zip to a public https server (http didn't work for me)
5. Clear application data
6. Launch the app, use the URL of your public https server hosting the bootstrap zip when prompted

### DIFF
--- a/modules/environment/login/default.nix
+++ b/modules/environment/login/default.nix
@@ -87,8 +87,8 @@ in
       prootStatic =
         let
           crossCompiledPaths = {
-            aarch64-linux = "/nix/store/7qd99m1w65x2vgqg453nd70y60sm3kay-proot-termux-static-aarch64-unknown-linux-android-unstable-2024-05-04";
-            x86_64-linux = "/nix/store/pakj3svvw84rhkzdc6211yhc2cgvc21f-proot-termux-static-x86_64-unknown-linux-android-unstable-2024-05-04";
+            aarch64-linux = "/nix/store/dvf2ck9bkw7yyrlkjk87xz1anaxsgrd6-proot-termux-static-aarch64-unknown-linux-android-unstable-2026-02-20";
+            x86_64-linux = "/nix/store/w7fksni07fhss7hzkyf4knn5171prj88-proot-termux-static-x86_64-unknown-linux-android-unstable-2026-02-20";
           };
         in
         "${crossCompiledPaths.${targetSystem}}";
--- a/pkgs/bootstrap.nix
+++ b/pkgs/bootstrap.nix
@@ -4,11 +4,13 @@
 
 runCommand "bootstrap" { } ''
   mkdir --parents $out/{.l2s,bin,dev/shm,etc,root,tmp,usr/{bin,lib}}
-  mkdir --parents $out/nix/var/nix/{profiles,gcroots}/per-user/nix-on-droid
+  mkdir --parents $out/nix
 
   cp --recursive ${nixDirectory}/store $out/nix/store
   cp --recursive ${nixDirectory}/var $out/nix/var
   chmod --recursive u+w $out/nix
+
+  mkdir --parents $out/nix/var/nix/{profiles,gcroots}/per-user/nix-on-droid
 
   ln --symbolic ${initialPackageInfo.bash}/bin/sh $out/bin/sh
 
--- a/pkgs/proot-termux/default.nix
+++ b/pkgs/proot-termux/default.nix
@@ -9,13 +9,13 @@
 
 stdenv.mkDerivation {
   pname = "proot-termux";
-  version = "unstable-2024-05-04";
+  version = "unstable-2026-02-20";
 
   src = fetchFromGitHub {
     repo = "proot";
     owner = "termux";
-    rev = "60485d2646c1e09105099772da4a20deda8d020d";
-    sha256 = "sha256-zHFPiL3ywZa8yzZa600BpoE+zuRipw2GNJrt3/Dy+/E=";
+    rev = "ab2e3464d04483b98a0614b470f3f8950d5a6468";
+    hash = "sha256-TMYkLmk+NnYcqJKF6RSOkN4S8AI5+HaNcgZZe/5E0vI=";
   };
 
   # ashmem.h is rather small, our needs are even smaller, so just define these:


newAM:
I did more testing, and eventually this broke.  I'm not sure if that was from restarting the app, or one of my rebuilds, or if this fix never really worked in the first place...

I think it's a good idea to update proot anyway, even if this isn't a complete fix.

Edit: Found the problem, I pointed my `nix-on-droid` flake input to this PR, but I needed to point it to a flake input that sets the `/nix/store` path in `modules/environment/login/default.nix`.  I didn't do that in this pull-request because the store path will change when built by a maintainer.

fdietze:
I tested this as well and it works!

verdict from my agent:
> Tested #529 on a **Pixel 10 Pro (Android 16)** — it fixes the tty problem (#515) for us. With the current proot, newer-glibc binaries (bash 5.3, zsh, vim, our AI agents) get `isatty(0)=false`; with this PR's proot they work again (arrow keys, prompts, TUIs). It let us drop several workarounds we'd accumulated: a pinned old bash, a pinned old Nix (for the same bug surfacing as #495 "pseudoterminal attributes"), and a static prompt. +1 to merge.
>
> One thing we still had to work around to *use* a custom proot: `environment.files.prootStatic` is `readOnly`, so it can't be overridden via the option — we install the bumped proot from a `build.activationAfter` script instead. Making it overridable would make testing proot fixes like this much easier.

shelvacu:
> because the store path will change when built by a maintainer.

why do you say that? the proot derivation (unlike a nix-on-droid activation package) should be pure

newAM:
> > because the store path will change when built by a maintainer.
> 
> why do you say that? the proot derivation (unlike a nix-on-droid activation package) should be pure

In my experience builds in nixpkgs are rarely pure.  For example any rust package will change with the number of CPU cores to build (a nixpkgs problem, not a rust one).  This package may be pure, but I by default assume packages are not.

shelvacu:
This is not a fixed-output derivation, so the nix store path uses a hash of the inputs, not the outputs.

newAM:
> This is not a fixed-output derivation, so the nix store path uses a hash of the inputs, not the outputs.

Ah, I had a misunderstanding in the way the store hashes were calculated.  I updated the PR with the new `crossCompiledPaths`

t184256:
> Copy the proot builds to a public nix cache, put your cache URL and key in modules/environment/nix.nix

Something's very wrong, this step should not be necessary at all. If this step is necessary, then your proot is not in the initial bootstrap zipball. CI bootstrap error suggests the same.

shelvacu:
I'm trying to figure out what went wrong with the CI and I am very confused.

If you look at the prepare-n-o-d step, the [file it produced](https://github.com/nix-community/nix-on-droid/actions/runs/27985285497/job/82825162196?pr=529#step:7:33) contains bootstrap-x86_64.zip, and that zip contains `/nix/store/w7fksni07fhss7hzkyf4knn5171prj88-proot-termux-static-x86_64-unknown-linux-android-unstable-2026-02-20` as it should. But if you look at the [screenshots](https://github.com/nix-community/nix-on-droid/actions/runs/27985285497/job/82826232982?pr=529#step:11:29), particularly `*-error.png`, it claims that exact same path does not exist.

shelvacu:
Of course, once I post the comment, I immediately figure it out lol

nix expects to have an accurate database of all paths in `/nix/store` at `/nix/var/nix/db/db.sqlite`.  the bootstrap derivation has a bug that puts this in the wrong spot:

https://github.com/nix-community/nix-on-droid/blob/55b6449b4582a4ba3ce712543c973360a026db7d/pkgs/bootstrap.nix#L6-L10

Instead of *creating* `$out/nix/var`, it copies *into* `$out/nix/var` because it already exists, which means the db file gets put at `$out/nix/var/var/nix/db/db.sqlite` which of course nix doesn't find so it doesn't know that the path exists.

When its able to fetch from cache, this isn't an issue.

DavidDTA:
> > Copy the proot builds to a public nix cache, put your cache URL and key in modules/environment/nix.nix
> 
> Something's very wrong, this step should not be necessary at all. If this step is necessary, then your proot is not in the initial bootstrap zipball. CI bootstrap error suggests the same.

I thought it would be necessary based on this section of the readme:

> Note: The proot binary is not built on the android device (NDK is required for building it, and it's not available on mobile platforms). The way we work around it is to push proot derivation to cachix. The current workaround is to hardcode the path to the wanted proot nix store path in modules/environment/login/default.nix. During evaluation time on the android device this store path will be downloaded from the binary cache (https://nix-on-droid.cachix.org/). This in return means the proot derivation has to be present there or in any other binary cache configured in the nix.conf on the device.

newAM:
> Instead of _creating_ `$out/nix/var`, it copies _into_ `$out/nix/var` because it already exists, which means the db file gets put at `$out/nix/var/var/nix/db/db.sqlite` which of course nix doesn't find so it doesn't know that the path exists.
> 
> When its able to fetch from cache, this isn't an issue.

Good catch!  I pushed another commit to this PR to fix that and (hopefully) get CI passing.

t184256:
Thank you very much.

t184256 (approved):

PR #533, opened 2026-05-09
modules: add programs.ssh

newAM:


newAM:
Finally had time to test this, it works as I intended.  I believe #529 will fix the CI failure.

t184256:
While I have tactical questions, the strategic one comes first: why do that if [home-manager has programs.ssh](https://nix-community.github.io/home-manager/options/home-manager/programs/ssh.html)?

newAM:
> While I have tactical questions, the strategic one comes first: why do that if [home-manager has programs.ssh](https://nix-community.github.io/home-manager/options/home-manager/programs/ssh.html)?

Yeah that's a fair point.  I'm lazy and wanted to import the same file that I use for NixOS, but delegating this to home-manager does make more sense for nix-on-droid as a project.

PR #540, opened 2026-07-05
modules: Add `runit` for service management

S0AndS0:
Long story short, Termux uses `runit` and (many hours/days) ago I thought it'd be easy/fun to get `runit` running with Nix on Droid x-]

Runit runs!... but, likely there are changes needed before this PR will be in an acceptable state

The added `modules/system/service/runit/README.md` provides a fully functional, albeit cringy, example of how I've got OpenSSH running with Runnit on an old Android device

The added `modules/system/service/runit/service.nix` file defines most of the options I could figure-out from spending quality time with related man-pages. Please check the related ReadMe's Notes/Warnings section for some notes and warnings

I'm not quite certain how best to link/load this module into the rest of the `config` name-space, and would much appreciate some assistance so no toes get stomped

### 2026-07-07

I'm not sure how to satisfy the CI/CD

Also am unsure if `runit-init` command will function as official docs advertise, I've not been able to make it happy

Also also am reasonably sure the following work-around is less than ideal for launching `runsvdir` automatically;

`flake.nix` (snippet)

```nix
  outputs = { self, home-manager, nixpkgs, nix-on-droid }@inputs:
  {
    nixOnDroidConfigurations.default = nix-on-droid.lib.nixOnDroidConfiguration {
      modules = [
        (
          { config, ... }:
          {
            home-manager = {
              extraSpecialArgs = {
                inherit pkgs;
                ## Forced even though it _should_ be set already
                osConfig = config;
              };

              # ... other configs for home-manager
            };
          }
        )
      ];
      # ... other configs for system
    };
  };
```

Home manager snippet

```nix
{
  config,
  lib,
  osConfig,
  ...
}:
{
  config.programs.bash = lib.mkIf config.programs.bash.enable {
    ## Extra commands to append to bashrc file
    bashrcExtra =
      ''
        # setup environment for non interactive sessions
        source "$HOME/.nix-profile/etc/profile.d/nix-on-droid-session-init.sh"
      ''
      + lib.optionalString osConfig.runit.enable ''
        # Who needs a daemon or full Init?!
        if ! pidof runsvdir >/dev/null 2>&1; then
          nohup runsvdir "${osConfig.runit.settings.environment.svdir}" >"${osConfig.runit.settings.environment.logdir}/runsvdir.log" 2>&1 &
          disown %1
        fi
      ''
      + ''
        # Commands that should be applied only for interactive shells.
        [[ $- == *i* ]] || return
      ''
      # ... other shell setup stuff
      ;
  };
}
```

### DIFF
--- a/modules/module-list.nix
+++ b/modules/module-list.nix
@@ -21,6 +21,7 @@
   ./environment/shell.nix
   ./home-manager.nix
   ./nixpkgs/options.nix
+  ./system/service/runit/service.nix
   ./terminal.nix
   ./time.nix
   ./upgrade.nix
--- a/modules/system/service/runit/README.md
+++ b/modules/system/service/runit/README.md
@@ -1,0 +1,273 @@
+# Runit for Nix on Android
+
+Once services are configured correctly the following command may be used to start all;
+
+```bash
+runsvdir ~/../usr/var/service
+```
+
+Within another session check process tree via;
+
+```bash
+ps xf
+```
+
+Each service `NAME` will create directories/files by default under `${config.build.installationDir}/var/service` with the following structure;
+
+- `./run` Required: used to start a service, defined by `runit.services.NAME.execute.run`
+- `./check` Optional: used to customize how service up/down state is checked, defined by `runit.services.NAME.execute.check`
+- `./conf` Not implemented
+- `./finish` Optional: called by runit when `./run` exits, defined by `runit.services.NAME.execute.finish`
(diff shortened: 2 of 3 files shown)


faretek1:
what about #203?

S0AndS0:
@faretek1 neat! Seems `supervisord` be a viable option for working around the issues that SystemD presents too

I'm using `runit` because that's what OG Termux suggests, it is what I started with within that ecosystem, and I think both `runit` and `superviosrd` can coexist as options for users to choose from here

